Issue #12, opened 2016-03-15
Allow jams to be cleared on reload


AACO:
in our ace settings use:
`
class ace_overheating_unJamOnreload {
  typeName = "BOOL";
  value = 1;
};
`
(moved from https://github.com/BourbonWarfare/bwmf/issues/77)

AACO:
resolved by PR #13

PR #13, opened 2016-03-16
Change some default ace settings

PabstMirror:
#12

These just change the defaults but do not force (lock) the settings, so mission makers can still set to whatever they want.

If we want we can remove these from the description.ext in BWMF

Ref http://ace3mod.com/wiki/framework/settings-framework.html#load-order

### DIFF
--- a/addons/aceSettings/config.cpp
+++ b/addons/aceSettings/config.cpp
@@ -1,0 +1,43 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"potato_core", "ace_laserPointer", "ace_map", "ace_zeus", "ace_hearing", "ace_overheating"};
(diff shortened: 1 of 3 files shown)


addons/aceSettings/config.cpp
```diff
@@ -0,0 +1,43 @@
...
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"potato_core", "ace_laserPointer", "ace_map", "ace_zeus", "ace_hearing", "ace_overheating"};
+        author[] = {"PabstMirror"};
+        authorUrl = "https://github.com/BourbonWarfare/POTATO";
+        VERSION_CONFIG;
+    };
+};
+
+
+// ACE Settings: (see http://forums.bourbonwarfare.site.nfoservers.com/viewtopic.php?f=8&t=2026)
```
AACO: Change to proper URL: http://forums.bourbonwarfare.com/viewtopic.php?f=8&t=2026

AACO:
Besides the outdated URL looks good :+1:

PR #15, opened 2016-03-23
scripting to make ares reinforcements happen natively on HC (or server if HC is not found)

AACO:
scripting to make ares reinforcements happen natively on HC (or server if HC is not found)


addons/reinforceToHC/functions/fnc_reinforcementsCreateUnits.sqf
```diff
@@ -0,0 +1,23 @@
+#include "script_component.hpp"
+
+// Define a scope so that inner calls can break to it if necessary to exit the logic while
+// still performing cleanup. The #define is visible in all module code that #includes this.
+#define MAIN_SCOPE_NAME "module_cleanup"
+scopeName MAIN_SCOPE_NAME;
```
AACO: The scope can probably get dropped here

addons/reinforceToHC/script_component.hpp
```diff
@@ -0,0 +1,16 @@
+#define COMPONENT reinforceToHC
+#include "\z\potato\addons\core\script_mod.hpp"
+
+#define DEBUG_MODE_FULL
```
AACO: Disable debug

addons/reinforceToHC/functions/fnc_getSpawnMachineId.sqf
```diff
@@ -0,0 +1,33 @@
...
+private _hcCounts = [];
+
+{
+    _hcIDs pushBack (owner _x);
+    _hcCounts pushBack 0;
+    nil
+} count _allHCs;
+
+{
+    private _index = _hcIDs find (groupOwner _x);
```
PabstMirror: btw, this load balancing wasn't actually working correctly with the other thing even when run on server, need to look into it

AACO: Does that mean it just spawns on the server, or doesn't work with multiple HCs?

PabstMirror: It just thinks the group count on all HCs is zero, so it will probably just spawn on the first HC all the time.
Not a problem for us now, and should be easy to backport the fix to this.

addons/reinforceToHC/functions/fnc_getSpawnMachineId.sqf
```diff
@@ -0,0 +1,33 @@
...
+    if (_index > 0) then {
```
AACO: This compare needs to be _index > -1. It's skipping index 0 (first HC) so it's not adding any of the units.

PabstMirror:
This should be good to go

PR #17, opened 2016-03-28
Add reset gear button to fix loadout issues

### LINKED ISSUE #9: Toolbox
Don't know how difficult this would be to implement, but maybe consider adding a basic toolbox we can use for fixing things like gear, radios, etc. That way code does not need to be run from debug.

For example:

Re-init everyone's gear/radios to fix mass nakedness/babel.
Able to select individuals to do the same. Either fix radio or gear or both.

PabstMirror:
Close  #9

### DIFF
--- a/addons/adminMenu/XEH_postInit.sqf
+++ b/addons/adminMenu/XEH_postInit.sqf
@@ -1,5 +1,25 @@
 #include "script_component.hpp"
 
+[QGVAR(resetGear), {
+    params ["_unit"];
+    diag_log text format ["[POTATO] Reseting Gear on %1 [%2]", (name _unit), _unit];
+    if ((isNull _unit) || {!alive _unit} || {!local _unit}) exitWith {};
+
+    removeAllWeapons _unit;
+    removeGoggles _unit;
+    removeHeadgear _unit;
+    removeVest _unit;
+    removeUniform _unit;
+    removeAllAssignedItems _unit;
+    removeBackpack _unit;
+
+    [{
+        diag_log text format ["[POTATO] Calling F_fnc_assignGearMan"];
+        _this call F_fnc_assignGearMan;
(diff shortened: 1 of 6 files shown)


addons/adminMenu/XEH_postInit.sqf
```diff
@@ -1,5 +1,25 @@
...
+    if ((isNull _unit) || {!alive _unit} || {!local _unit}) exitWith {};
+
+    removeAllWeapons _unit;
+    removeGoggles _unit;
+    removeHeadgear _unit;
+    removeVest _unit;
+    removeUniform _unit;
+    removeAllAssignedItems _unit;
+    removeBackpack _unit;
+
```
AACO: 1. we shouldn't be removing goggles here.
2. shouldn't the gear script do this anyways? (I know it won't double execute, but seems a bit wasteful)

addons/adminMenu/XEH_postInit.sqf
```diff
@@ -1,5 +1,25 @@
 #include "script_component.hpp"
 
+[QGVAR(resetGear), {
+    params ["_unit"];
+    diag_log text format ["[POTATO] Reseting Gear on %1 [%2]", (name _unit), _unit];
+    if ((isNull _unit) || {!alive _unit} || {!local _unit}) exitWith {};
```
AACO: The way we're executing this event, none of the above _shouldn't_ happen. We should log an error message on the exit.

AACO:
With the exception of my two minor comments, looks good to go.

addons/adminMenu/XEH_postInit.sqf
PabstMirror: Inventories can become deynced if you remove and re-add a uniform in a single frame, so there is a delay between the cleaning and re-adding.

PR #18, opened 2016-03-31
Hc extention

PabstMirror:


AACO:
The transfer to HC stuff doesn't quite work yet, so if we want to merge this ASAP, we should remove the unit and comment out the XEH for it.

PR #19, opened 2016-04-01
Parachute/dismounts

PabstMirror:


addons/zeusHC/functions/fnc_getSpawnMachineId.sqf
```diff
@@ -1,6 +1,8 @@
 #include "script_component.hpp"
 TRACE_1("Params",_this);
-params [["_allHCs", entities "HeadlessClient_F"]];
```
AACO: This will remove some cycle saving I did in the transfer to HC stuff.
https://github.com/BourbonWarfare/POTATO/blob/parachuteDismounts/addons/zeusHC/functions/fnc_transferGroupsToHC.sqf#L24
So it doesn't look for new entities on every new call within it's run.

addons/zeusHC/functions/fnc_createEntityDismounts.sqf
```diff
@@ -0,0 +1,81 @@
...
+#include "script_component.hpp"
+
+params ["_attachedVehicle", "_newGroup", "_createUnits", "_placerOwner"];
+TRACE_4("params",_attachedVehicle, _newGroup, _createUnits,_placerOwner);
+
+
+//Exit if bad conditions and send hint back to orginal module placer:
+if (isNull _attachedVehicle) exitWith {
+    diag_log text format ["[POTATO] - Attachable not attached"];
+    ["Module needs to be attached on a vehicle"] remoteExecCall [QACEFUNC(common,displayTextStructured), _placerOwner];
```
AACO: Maybe use bis_fnc_showCuratorFeedbackMessage instead? It doesn't matter to much, as long as the Zeus gets the feedback..

addons/zeusHC/functions/fnc_createEntityDismounts.sqf
```diff
@@ -0,0 +1,81 @@
...
+if ((!(_attachedVehicle isKindOf "Air")) && {(vectorMagnitude velocity _attachedVehicle) > 5}) exitWith {
+    diag_log text format ["[POTATO] - Ground veh too fast"];
+    ["Vehicle moving too fast"] remoteExecCall [QACEFUNC(common,displayTextStructured), _placerOwner];
+};
+
+
+private _cargoSelection = getText (configFile >> "CfgVehicles" >> (typeOf _attachedVehicle) >> "memoryPointsGetInCargo");
+if (_cargoSelection == "") then {
+    private _cargoSelection = getArray (configFile >> "CfgVehicles" >> (typeOf _attachedVehicle) >> "memoryPointsGetInCargo");
+    if (_cargoSelection isEqualTo []) exitWith {};
```
AACO: We should show a feedback message here, telling Zeus the vehicle doesn't support this feature.

addons/zeusHC/functions/fnc_createEntityDismounts.sqf
```diff
@@ -0,0 +1,81 @@
...
+            private _posASL = getPosASL _unit;
+            private _vel = velocity _unit;
+            _vel set [2, ((_vel select 2) + (0 max (-0.3 * (_vel select 2))))];
+
+            private _chute = "NonSteerable_Parachute_F" createVehicle _posASL;
+            TRACE_2("",_unit,_chute);
+
+            _unit moveInAny _chute;
+            //make sure the idiots don't get out
+            [{(_this select 0) moveInAny (_this select 1)}, [_unit, _chute], 1] call CBA_fnc_waitAndExecute;
```
AACO: This seems to be just a secondary call for the unit to move in the parachute (assuming they "move in"). Maybe loop this with a sleep until the outer loop's conditions are met?

addons/zeusHC/functions/fnc_createEntityDismounts.sqf
PabstMirror: It will still work with the fallback `_modelOffset = [0,-8,0];`, this just exits the if scope

addons/zeusHC/functions/fnc_getSpawnMachineId.sqf
PabstMirror: SQF can be pretty dumb and won't do things like short-circuit eval unless you make it.
Even if you pass in a value for `_allHCs` in the params, arma is still going to evaulate the expression for the default value.

Here's an example:
```
getEntities = {
    systemChat "computing getEntities"; 
    [1]
};

[[5]] call {
    params [["_test", call getEntities]];
    systemChat str _test
}
```
Will output "computing getEntites - [5]".

So this is no worse than before

addons/zeusHC/functions/fnc_createEntityDismounts.sqf
PabstMirror: They move in fine, but there's like a 5% chance the guy will think it's a good idea to get out a few frames later.  The delayed double move in seems to fix it.

PR #20, opened 2016-04-06
Boost ACRE's PRC-343 power

PabstMirror:
Boost 343 power output to 500mW (from 100mW)


AACO:
Lets wait to see how tonight goes without terlos before we jump to this.

PR #49, opened 2016-06-06
Should fix zeus list population time

AACO:


PabstMirror:
What problem does this fix?

IIRC I used `allPlayers` because it should give us compatibility with dead units in spectator.

AACO:
allPlayers seems to not fully synchronize until 2-3 minutes into the game (so if we need to give a player Zeus for setup/whatever it adds a large delay). Since we're using the "name" command, we need to filter by alive players. Also the way A3 handles dead players, they can't use Zeus anyways (unless we transferred them to an alive player/game logic)

PR #51, opened 2016-06-15
Added potato ratio calculator

AACO:
So, I recreated the old ratio calculator from the ground up.

It should:
- [x] Be significantly more resilient
- [x] Allow mission makers to define ratios in the description.ext file to pre-populate the ratio field
~~Allow mission makers to have civilians included in the ratio~~ RIP dreams
- [x] Allow users to set a player count override
- [x] Easier to maintain if an issue comes about

Ratio prefill explanation:
I'd love to stick with the description.ext approach, but that information is not available to clients at that MP screen, to work around this Pabst suggested we just parse the description line, so that's what we now do.

Prefill order:
BlueFor:OpFor:Indy
So a 3 BlueFor to 2 Indy ratio would simply be "3:2". A 4 OpFor to 3 Indy ratio would simply be "4:2".

Note:
The ratio must be the first item on the mission line, it must be only numbers, and any missing values will be returned as 1.

Example mission description:
```
3:2 other mission info here
```

Display:
http://i.imgur.com/03RpO8X.png

Question:
~~Right now I have civilians setup to be calculated in the ratio system, is that something we want/need? Should we add an option to disable it?~~ addressed


AACO:
(also, I never want to write code for whatever fucking crazy environment this is in...)

zerithsmash:
Being able to set ratios in the MMing process is an amazing idea. :+1: 

I don't know where you would put the option to disable counting civilians, but I think it helps if it omits slotted civ slots for those types of missions so there isn't any confusion.

AACO:
Pull message updated, should be ready to merge

AACO:
@PabstMirror should we revert the 3DEN changed in mission modules and core?

PabstMirror:
Those are fine, we'll want it if we ever add anything else to threeden.

PR #54, opened 2016-06-22
Markers

AACO:
**Ready to go bby**

Moving the marker system out of the framework, to the mod side. This new system should be easily configurable via ace settings, and through 3DEN attributes, instead of relying on name matching.

Unit marker setup: http://i.imgur.com/FMAu4Ti.jpg
Group marker setup: http://i.imgur.com/NPAjieV.jpg


AACO:
Ready to merge, description updated with pictures.

AACO:
Note will break the current marker aspect of the BWMF respawn script, should implement a stop gap solution when updating the BWMF to use these markers

PR #56, opened 2016-06-24
Potato color teams

AACO:
Tested locally and on a local dedicated server, seems to work okay, can't really test multi-player though.


PabstMirror:
I think APEX will actually do fireteam color syncing correctly,
When that happens we should be able to simplfy and lose the initXEH.

So, we might want to wait for 1.62 if the 3den expression is changing.
I will try to do some testing and see if it actually works.

PR #62, opened 2016-07-03
Force STHUD to be configured through ACE settings instead of the built in UI

### LINKED ISSUE #61: Add "sthud/ace" compat to allow setting "default" hud setup
* Allow you to turn on/off compass on the hud
* Allow you to set the display as (FULL|COMPASS|TEAM MEMBERS|OFF)

Is there any actual interest in this?

AACO:
Disable STHUD menu
Move settings to client settable ACE Settings (allows mission makers to force HUD modes)
Resolves #61

Note to @PabstMirror , if you want to build this as an ACE Compat PBO, you should just need to localize the setting strings. I wanted to entirely disable the STHUD menu's keybinding but the way they have it set up doesn't provide a good way to go that route. We could always just drop the STHUD menu PBO if this gets added.

More general note: I should probably set something up similarly for Apex HUD, but I still got some time there.

Assigned to @zerithsmash and @PabstMirror to discuss if this is something we even want to do.

### DIFF
--- a/addons/sthud/ACE_Settings.hpp
+++ b/addons/sthud/ACE_Settings.hpp
@@ -1,0 +1,17 @@
+class ACE_Settings {
+    class GVAR(stHudMode) {
+        category = "ST HUD";
+        displayName = "ST HUD Mode";
+        typeName = "SCALAR";
+        value = 3;
+        values[] = {"OFF", "HUD Only", "Names Only", "Normal"};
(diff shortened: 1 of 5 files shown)


zerithsmash:
Looks ok to me. I didn't see any obvious errors in code, but might be good to have pabst take a quick look.

AACO:
¯\_(ツ)_/¯

PR #63, opened 2016-07-03
Add semi-transparent background to apex squad radar

PabstMirror:
Add background to Apex squad radar so it looks more like what we are used to.
image


addons/miscFixes/config.cpp
```diff
@@ -82,3 +82,11 @@ class CfgWeapons {
         };
     };
 };
+
+class CfgInGameUI {
+    class SquadRadar {
+        //78-64-40
+        // backgroundTexture = "";
+        compassBackgroundTexture = QUOTE(PATHTOF(ui\squadRadarBackground_Darker_ca.paa));
```
AACO: Have you tried this at different resolutions/aspect rations? I'm not sure how using a texture will work.

PabstMirror: In vanilla it's a texture with the 3 rings and compass dirs.  I just filled in the inside area

AACO:
Has BI just dropped support for this?

PR #86, opened 2016-09-03
Fix markers being shown for empty groups

PabstMirror:
The game seems to still create all groups even if they are not slotted.
So markers is adding markers for all groups and draws icons at [0,0,0]
This just moves them out of the way.

Example, running bwmf with no AI and just me slotted:


AACO:
Looks good, is this something we have to consider anywhere else where? (Any sort of side-group limits?)

PabstMirror:
I really not sure if this will be a problem; adding MSV will put us over 144 which might be the limit?

PR #90, opened 2016-09-26
Team colors fix

AACO:
Should fix team colors, only tested locally though, a few person MP test should really be executed before merging.

### DIFF
--- a/addons/teamColors/functions/fnc_setColor.sqf
+++ b/addons/teamColors/functions/fnc_setColor.sqf
@@ -11,7 +11,7 @@
  *
  * Public: No
  */
-#define DEBUG_MODE_FULL
+
 #include "script_component.hpp"
 
 TRACE_1("Params",_this);
--- a/addons/teamColors/functions/fnc_setColor.sqf
+++ b/addons/teamColors/functions/fnc_setColor.sqf
@@ -15,7 +15,4 @@
 
 TRACE_1("Params",_this);
 
-[{
-    params ["_unit", "_fireteam"];
-    _unit assignTeam _fireteam;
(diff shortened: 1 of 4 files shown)


PabstMirror:
I have no clue why this broke
event is still there in cba

BIKI says global/global, but I remember it being a little weird when I was testing 1.62

AACO:
It looks like the event is registered postInit, while this can be called during init (the waitAndExecute _should_ push us past post init), but I'm guessing that it's some strange race issue.

addons/teamColors/functions/fnc_setColor.sqf
```diff
@@ -15,8 +16,8 @@
...
 TRACE_1("Params",_this);
 
-[{
-    params ["_unit", "_fireteam"];
-    TRACE_2("setting fireteam",_unit,_fireteam);
-    ["CBA_teamColorChanged", [_unit, _fireteam], [_unit]] call CBA_fnc_targetEvent;
-}, _this, 1] call CBA_fnc_waitAndExecute;
+[
+    {(_this select 0) assignTeam (_this select 1);},
+    {!isNull player},
```
PabstMirror: I think the order is reversed, `[condition,statement,args]`
But this also runs on dedi, so player will be null there

AACO: Good call, I'm still playing with it a bit to do testing, assignTeam needs to be called when the player is finished initializing, so we might be better off checking unit locality then waiting for player initialization.

AACO:
Seems to work on dedicated with AI, worth a try......

PR #91, opened 2016-09-27
Respawn / Zeus support for radios

PabstMirror:
Adds support for `ace_player` so `FUNC(setupPlayer)` will rerun on each change

Respawns, including side change respawns, should work
As well as zeus RCing units.

### DIFF
--- a/addons/radios/XEH_postInit.sqf
+++ b/addons/radios/XEH_postInit.sqf
@@ -5,16 +5,12 @@ if (isServer) then {
     [] call FUNC(setAcreGlobalSettings);
     [] call FUNC(setAcreSideSettings);
 };
- */
+*/
 
 // Remove by october
-if (isNil "acre_api_fnc_ignoreAntennaDirection") then {
-    INFO_1("Setting omnidirectional for acre pre-904",true);
-    acre_sys_signal_omnidirectionalRadios = true;
-} else {
-    INFO_1("Setting omnidirectional for acre post-904",true);
-    [true] call acre_api_fnc_ignoreAntennaDirection;
-};
+INFO_1("Setting omnidirectional for acre post-904",true);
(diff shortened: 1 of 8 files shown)


addons/radios/XEH_postInit.sqf
```diff
@@ -33,58 +29,11 @@ if (isNil "acre_api_fnc_ignoreAntennaDirection") then {
...
-                GVAR(initState) = 4;
-                _return = true;
-            };
-            _return
-        },
-        {
-            INFO_2("%1 - [InitState %2] DONE", diag_frameNo, GVAR(initState));
-        }, []] call CBA_fnc_waitUntilAndExecute;
+        
+        ["unit", {_this call FUNC(setupPlayer)}, true] call CBA_fnc_addPlayerEventHandler;
```
AACO: Will CBA's `missionNamespace getVariable QGVAR(oldUnit);` be correct at this point? It should be fine for zeus/reinforcements though.

PabstMirror: By default it will be `objNull`, which won't match and will add the briefing
Setting the object instead of a bool is just a respawn safe way of setting a variable on something.

PabstMirror: https://github.com/CBATeam/CBA_A3/blob/master/addons/events/fnc_addPlayerEventHandler.sqf#L44
```
if (_applyRetroactively && {!isNull (missionNamespace getVariable [QGVAR(oldUnit), objNull])}) then {
        [GVAR(oldUnit), objNull] call _function;
    };
```
If it's null it won't run, but will next change.

AACO: As long as that event gets triggered along the normal spawn/jip path 👍

AACO:
Happy path looks okay

PR #93, opened 2016-10-06
Increase illumination from vanilla and rhs flares

PabstMirror:


AACO:
Shouldn't this be (mostly) addressed by https://github.com/acemod/ACE3/pull/4506 ?

PabstMirror:
- This PR will work for us right away, with no change to loadouts
- The ACE PR won't touch RHS flares, which we are going to want to make it symetrical
- I'm not sure if that ace PR is the right way, I'm also leaning towards adding new magazines. If I can figure out how to do it.

AACO:
As to point 3, good fucking luck, maybe the Tanoa changes fixed it, but I tried ~3 hours to get it working for the UGLs to very limited success.

AACO:
One last thing before merging this, should we add this to the BWA3 gustav illumination rounds? It might be too niche to worry about, just wanted to bring it up.

PR #95, opened 2016-10-09
SMAW playability tweaks

AACO:
Should lighten up the SMAW enough to make it a viable replacement for BluFor MAT teams. The gunner should be able to hold 5 magazines (one loaded) with the spotter holding four spare mags.


zerithsmash:
Fine with this.

PR #96, opened 2016-10-12
Add throwable potato "grenade"

PabstMirror:


AACO:
Can vs Should, I'll let @zerithsmash decide

zerithsmash:
If it leads to more pabo missions...should?

PR #98, opened 2016-10-16
Remove group ID fix, now in CBA

AACO:
Should now be handled in CBA: https://github.com/CBATeam/CBA_A3/pull/508

Note: this may need to be fixed/converted in the framework first?
Note2: https://github.com/BourbonWarfare/bwmf/pull/108 should be merged well before this item.


AACO:
Instead of completely removing the function, we should shim it, and just have it call CBA's method, to be removed many releases later.

PR #102, opened 2016-10-24
Side marker updates

### LINKED ISSUE #82: Allow user to choose side specific marker icon
Allow user to choose side specific marker icon

AACO:
- Adds a (yet unused) core function to easily convert a string or a number to a side (really thought I was going to need it)
- Updates the side specific marker to have more options, hopefully easier to set. (Resolves #82 )


AACO:
Good fixes. I guess object arguments are globally executed, but object attributes are not?

PabstMirror:
I think arguments still only run on server, they just have the global bit set to true automaticly.
Here's the sqm for the old format:
```
				class Attribute1
				{
					property = "potato_missionModules_sideMarker_side";
					expression = "_this setVariable ['side',_value,true];";
					class Value
					{
						class data
						{
							class type
							{
								type[] = {"SCALAR"};
							};
							value = 0.0;
						};
					};
				};
```

AACO:
Ahh, so it's done behind the scenes when saving the SQM, makes sense.

Issue #94, opened 2016-10-07
Potato assignGear vs ArmA gear initialization race condition.


AACO:
It appears that, upon JIPing assignGear races with the normal gear initialization. Due to this race, about 50-80% of JIPs get the vanilla gear, and we get the occasional ACRE sync issue.

AACO:
Can we change https://github.com/BourbonWarfare/POTATO/blob/master/addons/assignGear/CfgEventHandlers.hpp#L13 to post init? and delay it by a frame? Right now we have issues with both JIPs and units that try to use the `F_Gear` override on their init scripting.

AACO:
Even with a frame delay JIPs are still having issues, do we want to try and add a larger delay and hope for the best, or is there any better way to handle this?

PabstMirror:
Shame this is impossible to reproduce localy, 
I've got an idea, I'll try to make a PR later today

AACO:
Semi resolved by #113 ... we could probably use a better solution though...

AACO:
Would the be better executed on the local level with a global cache?

PR #106, opened 2016-10-30
Assign Gear refactor/update

AACO:
hooo boy, major update to assign gear.
- Added a frame delay to the assign gear (man and vehicle) scripts to help with the `F_Gear` and JIP issues (ref #94 ).  `F_Gear` seems to work much better, JIPing remains to be seen.
- Add `handgunAttachments` to the available loadout fields (mainly for competitionist sake, since we have the launcher attachments as well)
- Break `getLoadoutFromConfig` into a few helper functions to A. reduce SQF calls (define functions once, instead of each run) and B. to hopefully helpfully break down some of the 'mega code' into more readable parcels.
- Add more debugging around assignGearMan to hopefully give better benchmarking.
- Add ability to have ambiguous loadouts. The TL;DR of this feature is to have one loadout (german, british, whatever) usable for all sides/factions.
- Don't run the magnified optic check on launchers (prevents a lot of MAT issues honestly).
- Set the class event handlers through scripting, instead of through configuration

Running on my computer (local dedicated server, local client, value in seconds):
`1:16:10 "[POTATO-assignGear] - assignGearMan ran 588 times, with an average run time of 0.00462764"`
4ms be call ain't bad.

@PabstMirror can you take a read through? I did local/dedicated testing and it seemed fine.
@zerithsmash tagging you just because.

### DIFF
--- a/addons/assignGear/CfgEventHandlers.hpp
+++ b/addons/assignGear/CfgEventHandlers.hpp
@@ -12,21 +12,21 @@ class Extended_PreInit_EventHandlers {
 
 class Extended_Init_EventHandlers {
     class CAManBase {
-        class ADDON { serverInit = QUOTE(_this call FUNC(assignGearMan)); };
+        class ADDON { serverInit = QUOTE([ARR_2(FUNC(assignGearMan),_this)] call CBA_fnc_execNextFrame); };
     };
     class Car {
-        class ADDON { serverInit  = QUOTE([ARR_2((_this select 0), 'Car')] call FUNC(assignGearVehicle);); };
+        class ADDON { serverInit = QUOTE([ARR_2(FUNC(assignGearVehicle),[ARR_2(_this select 0, 'Car')])] call CBA_fnc_execNextFrame); };
     };
     class Tank {
-        class ADDON { serverInit  = QUOTE([ARR_2((_this select 0), 'Tank')] call FUNC(assignGearVehicle);); };
+        class ADDON { serverInit = QUOTE([ARR_2(FUNC(assignGearVehicle),[ARR_2(_this select 0, 'Tank')])] call CBA_fnc_execNextFrame); };
     };
     class Helicopter {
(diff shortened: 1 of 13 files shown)


AACO:
Example of what would be in the CfgLoadouts.hpp file:
```
  prefixes[] = {"b_", "o_", "i_"};

  //Shared items
  #include "Loadouts\common.hpp" // DO NOT COMMENT OUT, WILL BREAK EVERYTHING

  //Only include one hpp per faction; use (//) to comment out other files
  class blu_f {
	#include "Loadouts\ger_g36_fleck.hpp"
  };
  class opf_f {
	#include "Loadouts\ger_g36_fleck.hpp"
  };
  class ind_f {
	#include "Loadouts\ger_g36_fleck.hpp"
  };
```
the prefixes line is in there to make sure we don't strip the prefixes on non-compatible loadout files.

PabstMirror:
looks great as far as added functionality

I am worried about the delay, as I had problems with it causing gear desync / acre issues
since then the command might have improved
it's hard to test as session conditions will always be different than small scale
e.g. - I can't reproduce the current jip issue

but if it fails, worst case would be just doing a mass gear reset via console

AACO:
The JIP issue will be very hard to nail down, but I know the F_Gear variable doesn't get set before the scrpt is run, locally or dedicated. If it's just ACRE gear desync we can just reset that message.

addons/assignGear/CfgEventHandlers.hpp
```diff
@@ -12,21 +12,21 @@ class Extended_PreInit_EventHandlers {
 
 class Extended_Init_EventHandlers {
```
AACO: Should these be added through code using `CBA_fnc_addClassEventHandler` ? instead of pre-registering these event handlers?

AACO:
On 1.0.17 release, keep an eye out for ACRE ID issues.

PR #109, opened 2016-11-25
Update config.cpp

Spartachris:
Achilles is backwards compatible with the patches made in potato with this change. Needs testing on Dedicated Server.


AACO:
Will be address by ace, instead of the Achilles Ares iteration.

PR #113, opened 2016-12-07
Bandaid to auto reset gear on failure

PabstMirror:


addons/assignGear/functions/fnc_assignGearMan.sqf
```diff
@@ -60,18 +59,22 @@ if (isText (_path >> "init")) then {
...
+    private _arrayUniform = (_loadoutArray select 3) param [0, ""];
+    private _arrayVest = (_loadoutArray select 4) param [0, ""];
+    INFO_4("Checking loadout: Uniform [%1-%2] Vest [%3-%4]",_arrayUniform,uniform _unit,_arrayVest,vest _unit);
+
+    if ((_arrayUniform != (uniform _unit)) || {_arrayVest != (vest _unit)}) then {
+        ERROR_2("Mismatch [%1] [%2]",name _unit, typeOf _unit);
+        _unit setUnitLoadout _loadoutArray;
+        ["potato_adminMsg", [(format ["Auto-reset gear on %1", name _unit]), "Server"]] call CBA_fnc_globalEvent;
+    };
+}, [_unit, _loadoutArray], 15] call CBA_fnc_waitAndExecute
```
AACO: Do we have to worry about this blowing upt the waitAndExecute array if we have a high player/AI count?

addons/assignGear/functions/fnc_assignGearMan.sqf
```diff
@@ -60,18 +59,22 @@ if (isText (_path >> "init")) then {
...
+// Stupid hack to fix gear automaticly until we can sort out why setUnitLoadout ocasionly fails on JIPs
+[{
+    params ["_unit", "_loadoutArray"];
+    if (isNull _unit) exitWith {};
+    
+    private _arrayUniform = (_loadoutArray select 3) param [0, ""];
+    private _arrayVest = (_loadoutArray select 4) param [0, ""];
+    INFO_4("Checking loadout: Uniform [%1-%2] Vest [%3-%4]",_arrayUniform,uniform _unit,_arrayVest,vest _unit);
+
+    if ((_arrayUniform != (uniform _unit)) || {_arrayVest != (vest _unit)}) then {
```
PabstMirror: It's not perfect, but most of the time they'll be different

addons/assignGear/functions/fnc_assignGearMan.sqf
PabstMirror: waitAndExecute is sorted and only 1 will be checked each frame, so it's not too bad
But it also makes sorting more expensive... either way it shouldn't matter to much

PR #114, opened 2016-12-09
Optix

AACO:
Usage example: http://i.imgur.com/aVdKMJD.gif
Configuration:
`allowChangeableOptics = 1;`
in CfgLoadouts.hpp
```
class Soldier_F {// rifleman
  opticChoices[] = {"optic_Aco","optic_Holosight_blk_F"};
};
```
in the loadout itself

Features:
-Choosable optic system driven by ACE self interaction and out loadout configs
-Only accessible during safe start

Questions ( @zerithsmash ):
-Is this something we want as a group?
-Do we need to do the magnified optics check on these optics? (Not currently implemented, should be easy though)
-Do we need to handle JIPs?
-Do we need to handle respawns?


Alex6511:
Would really like this for the new loadouts coming, it would make handling a lot of oddities easier. Probably a good idea to do magnified optics checks on them just be to be safe. IMO JIPs and Respawns should just be given a default optic if safestart has ended.

mod.cpp
```diff
@@ -1,4 +1,4 @@
-name = "Potato 0.0.0";
+name = "Potato 1.0.18";
```
PabstMirror: This gets set automatically by build.py, so 0.0.0 is fine

AACO:
@PabstMirror is there a better way to handle the inventory check?

PR #115, opened 2016-12-11
Add dumb NLAW, add 65/762 tracer mags

AACO:
Adds:
-Dumb fire only NLAW
-Mk200 and Zafir tracer mags in red/green/yellow

_Should_ help creating vanilla loadouts


zerithsmash:
Sure

PR #117, opened 2016-12-14
Add recruit ID tool

AACO:
- [x] Move create rsc title functionality to core
- [x] Update/cleanup safestart (and fix a bug along the way)
- [x] Implement message to display to members about recruits during safe start
- [x] Implement a recruit 3D tag to display to members during safe start
- [x] Update/cleanup comments

Notes:
- Driven by the squad.xml, so members w/o tags will be ID as recruits (and should be appropriately heckled)
- Will be able to turn off the message and/or the tags via ACE Settings UI

Message example:
message

Tag example:
http://i.imgur.com/7koi6fq.gif

Question(s):
-Are the tags baller?

Tags:
@zerithsmash


zerithsmash:
I likey. Let the heckling begin.

PR #118, opened 2016-12-27
Add debug menu for authoried techs, minor cleanup

AACO:
Add the escape menu debug menu to any user who is authorized by the `potato_core_fnc_isTech` command. Also some minor cleanup in core/briefing.

Should allow authorized users to easily fix minor issues or use the splendid cam to take some sweet screenies.

Question to the admins: is it too much power vs responsibilities?


addons/core/config.cpp
```diff
@@ -33,5 +33,3 @@ class CfgMods {
         description = "";
     };
 };
-
-// JOKE
```
PabstMirror: I'm not sure about this line being removed.

PR #119, opened 2016-12-31
Predator Vision

AACO:
http://i.imgur.com/cvwvv62.gifv


AACO:
In hopes that it gets revied

PR #122, opened 2017-02-03
Banzi Attack Module

PabstMirror:


addons/missionModules/functions/fnc_zeus_banziCharge.sqf
```diff
@@ -0,0 +1,15 @@
+#include "script_component.hpp"
+
+params ["_logic"];
+
+if (!local _logic) exitWith {};
+private _unit = _logic getVariable ["bis_fnc_curatorAttachObject_object",objnull];
```
AACO: Do we want to use the dismount workaround logic here? `bis_fnc_curatorAttachObject_object` should be accurate, but may not be....

addons/missionModules/CfgVehicles.hpp
```diff
@@ -98,4 +98,16 @@ class CfgVehicles {
...
         class Attributes {};
     };
+
+    class GVAR(zeus_banziCharge): Module_F {
+        author = QUOTE(PREFIX);
+        category = QUOTE(PREFIX);
+        scope = 1;
+        scopeCurator = 2;
+        isTriggerActivated = 0;
+        isGlobal = 2;
```
AACO: Why are we using persistent global here?

addons/missionModules/XEH_postInit.sqf
```diff
@@ -26,3 +26,21 @@ GVAR(ehAdded) = false;
...
+    TRACE_4("",behaviour _unit,combatMode _group,speedMode _group,formation _group);
+    _group setBehaviour "AWARE";
+    _group setCombatMode "RED";
+    _group setSpeedMode "FULL";
+    _group setFormation "LINE";
+
+    {
+        TRACE_1("member",_x);
+        _x disableAI "AUTOCOMBAT";
+        _x disableAI "COVER";
```
AACO: We should disable `SUPPRESSION` as well. Maybe make courage to 1 as well?

addons/missionModules/functions/fnc_zeus_banziCharge.sqf
```diff
@@ -0,0 +1,15 @@
...
+TRACE_2("",_logic,_unit);
+
+if ((!isNull _unit) || {alive _unit} || {_unit isKindOf "CaManBase"}) then {
+    [QGVAR(banzi), [group _unit], [leader _unit]] call CBA_fnc_targetEvent;
```
AACO: Just curious, is there any advantage to using an event system here, instead of remoteexec-ing? If we don't plan on having hooks in other addons, it seems like we're using a lot of  piping for a simple call

addons/missionModules/CfgVehicles.hpp
PabstMirror: Module runs were local so we get the correct attach thing
Gets deleted after, so should be fine

addons/missionModules/functions/fnc_zeus_banziCharge.sqf
PabstMirror: should be fine because module is local to the zues

addons/missionModules/XEH_postInit.sqf
PabstMirror: I don't want to turn them into unstoppable robots
mainly just get them up and moving
we can tweak it later, or add a 2nd module with stronger effects

addons/missionModules/CfgVehicles.hpp
AACO: I mean, I agree on using a global scope, but I don't see the need to explicit mark it persistent

addons/missionModules/XEH_postInit.sqf
AACO: Not unstoppable... just inspired by the great emperor....

PR #129, opened 2017-03-08
Add gesture color based on map marker

AACO:
Simple code addition to set the map gesture to the color of the group's unit color. Should be nice for large scale briefings

Gesture Image


AACO:
It's technically a frame after init, but I did a bunch of dedicated testing without issue, although we'll see how it rolls with large scale MP

PR #138, opened 2017-03-21
Fire event on mission end

AACO:
- [ ] Tested/verified on local/dedicated servers


AACO:
Closing for now, will probably make a more agnostic setup

PR #143, opened 2017-04-03
Change marker init path

AACO:
Uses player group side
Uses add player event handler retroactive activation
Downside: the global event fires for each client on startup :\

### DIFF
--- a/addons/markers/functions/fnc_getSideArray.sqf
+++ b/addons/markers/functions/fnc_getSideArray.sqf
@@ -11,7 +11,7 @@
 #include "script_component.hpp"
 TRACE_1("Params",_this);
 
-private [["_side", sideLogic, [west]]];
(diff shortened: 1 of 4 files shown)


AACO:
This seems to be needed for second load of compressed mission (and at semi high player count, best guess: playerSide isn't correct)

addons/markers/functions/fnc_getSideArray.sqf
```diff
@@ -11,7 +11,9 @@
 #include "script_component.hpp"
 TRACE_1("Params",_this);
 
-switch (playerSide) do {
+private [["_side", sideLogic, [west]]];
```
PabstMirror: Should this be `params`?

AACO: yup

AACO:
Closing for a different solution

PR #150, opened 2017-05-07
Admin menu update for 1.70

AACO:
This really sucked to update, like almost as much as the ratio component.
_Should_ keep debug working like it used to by overriding BIS_fnc_isDebugConsoleAllowed


addons/adminMenu/functions/fnc_isDebugConsoleAllowed.sqf
```diff
@@ -0,0 +1,16 @@
+private _enableDebugConsole = [
+    "DebugConsole",
+    getMissionConfigValue ["enableDebugConsole", 0]
+] call (missionNamespace getVariable "BIS_fnc_getParamValue");
+
+ // Always allowed by mission config or mission param
+if (_enableDebugConsole isEqualTo 2 || {hasInterface && isServer}) exitWith { true };
+
+// In single player editor
+if (!isMultiplayer && { allDisplays find findDisplay 26 in [0, 1] || allDisplays find findDisplay 313 in [0, 1] }) exitWith { true };
```
PabstMirror: If it's not isMultiplayer , we should have already matched with the line above: `hasInterface && isServer`.
So this isn't necessary but won't hurt.
I guess `DebugConsole = 0` will be ignored, but I see no problem with that

addons/adminMenu/CfgFunctions.hpp
```diff
@@ -0,0 +1,12 @@
+class CfgFunctions {
+    class A3 {
+        tag = "BIS";
+        project = QUOTE(ADDON);
```
PabstMirror: Probably remove tag and project because those effect all of `A3` which I don't think we want to be messing with.

addons/adminMenu/functions/fnc_isDebugConsoleAllowed.sqf
AACO: Yeah, this file changed like 80 times due to restrictions to what I could access, I can simplify

addons/adminMenu/CfgFunctions.hpp
AACO: Yeah, it was leftover when it wasn't under the A3 class. Good catch.

PR #151, opened 2017-05-14
Lint cleanup

PabstMirror:
- Fix some BANZI stuff
- acco - could you double check `addons/radios/functions/fnc_channelControlLoad.sqf`
- everything else is minor stuff (e.g. var exists from a higher scope)

### DIFF
--- a/addons/missionModules/XEH_postInit.sqf
+++ b/addons/missionModules/XEH_postInit.sqf
@@ -31,7 +31,7 @@ GVAR(ehAdded) = false;
     params ["_group"];
     TRACE_1("banzi eh", _group);
 
-    TRACE_4("",behaviour _unit,combatMode _group,speedMode _group,formation _group);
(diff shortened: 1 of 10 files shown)


AACO:
Looks good, the channel control load condition you fixed should never actually trip and was just in there for strange-state security.

I'm surprised we didn't see bigger problems with _factory/_factoryLogic

AACO:
Do we want to set up build jobs to run some of these scripts on the merge requests?

PabstMirror:
A lot of them were setup like this, such that the vars still existed, just at an unexpected scope.
A:  
```
params ["_factoryLogic"];
[_factoryLogic] Call B;
```
B:
```
params ["_factory"];
_factoryLogic getVar .... //_factoryLogic existed, just from A's scope
````

I guess build checking tools might be nice, the linter is making rapid progress but still has some false positives right now.

PR #155, opened 2017-06-02
Deprecate no radar

PabstMirror:
After 1.70 changes this isn't really needed as much
Changes to a vehicle's radar will have to be done differently.

This still leaves the vehicles as available, so shouldn't break anything.
Can remove later.


AACO:
I'm all about deprecating r8, it did what we needed, but was a bit of a pain to maintain

PR #157, opened 2017-06-11
Assign gear tweaks

AACO:
Changes:
- Use core funcs when pulling information from the configs
- Move the XEHs from init to initPost (should have the frame delay built in now)
- Make gear apply to units on 3DEN load
- Make sure allowChangeableOptics is defined even if potato is off
- Add backpack validation to the gear watchdog
- Change the vehicle gear class lookup to scan through available faction/loadout combos before looking at the default loadout.

Questions:
Do we want to look at ALL factions under a side, instead of the limited scope (add some caching once found)? Would increase initial lookup time, but would not need to update this code if we expand our set of default factions?

### DIFF
--- a/addons/assignGear/XEH_preInit.sqf
+++ b/addons/assignGear/XEH_preInit.sqf
@@ -9,8 +9,6 @@ PREP_RECOMPILE_END;
 GVAR(usePotato) = [missionConfigFile >> "CfgLoadouts" >> "usePotato"] call CFUNC(getBool);
 
 if (GVAR(usePotato)) then {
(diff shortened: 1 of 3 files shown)


PabstMirror:
assignGear man has been super touchey about when it gets run
also need to think about how it effects acre setup
so probably want to test this on a dedicated

but overall looks good

AACO:
Good call. ACRE seemed fine, but initPost 100% failed for JIP testing (where init seems to always work on local dedicated servers)

PabstMirror:
but jips will very often have to run the fallback anyways

PR #160, opened 2017-07-14
Zeus Recon

PabstMirror:
Attempts to have zeus AI give contact reports much like you might get from real players.
Only reports what zeus AI have actually seen (instead of zues magically seeing all players)

Enemies reported with orange marks, showing composition and time last seen:
image

### DIFF
--- a/addons/zeusRecon/functions/fnc_reconStateMachine.sqf
+++ b/addons/zeusRecon/functions/fnc_reconStateMachine.sqf
@@ -1,0 +1,304 @@
+#include "script_component.hpp"
+
+#define RANGE_TO_GROUP 75
+#define TIME_KEEP_SHOWING_AFTER_LOSING_SIGHT 20
+#define ICON_HEIGHT_ABOVE_GROUND 20
+#define RANDOM_GROUP_COLLAPSE_TEST 0.01
+#define REPORT_ICON_MEDICAL "\A3\Ui_f\data\IGUI\RscIngameUI\RscUnitInfo\icon_cross_ca.paa"
+#define REPORT_PRIOITY_MEDICAL 3
+#define REPORT_ICON_RECON "\A3\Ui_F_Curator\Data\Logos\arma3_curator_eye_32_ca.paa"
+#define REPORT_PRIOITY_RECON 1
+
+switch (GVAR(state)) do {
+case (0): { // Start: Gather friendly units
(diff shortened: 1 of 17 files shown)


Alex6511:
👍

addons/zeusRecon/functions/fnc_zeusInterfaceOpenedEH.sqf
```diff
@@ -0,0 +1,43 @@
...
+    GVAR(friendlyUnits) = [];
+    GVAR(friendlyGroups) = [];
+    GVAR(friendlyReports) = [];
+
+    GVAR(addToCurator) = [];
+    GVAR(removeFromCurator) = [];
+
+    if (isNull getAssignedCuratorLogic player) then {systemChat "Unknown Curator?";};
+
+    addMissionEventHandler ["EachFrame", {
```
AACO: Any serious reason we're not using CBA's FSM or BIs FSM setups?

PabstMirror: CBA state machine is designed for objects.
this loops through both objects and general arrays

Main advantage is we can control the update rate, right now it's 5x each frame (probably higher than it needs to be)

addons/zeusRecon/functions/fnc_reconStateMachine.sqf
```diff
@@ -0,0 +1,304 @@
...
+#define RANGE_TO_GROUP 75
+#define TIME_KEEP_SHOWING_AFTER_LOSING_SIGHT 20
+#define ICON_HEIGHT_ABOVE_GROUND 20
+#define RANDOM_GROUP_COLLAPSE_TEST 0.01
+#define REPORT_ICON_MEDICAL "\A3\Ui_f\data\IGUI\RscIngameUI\RscUnitInfo\icon_cross_ca.paa"
+#define REPORT_PRIOITY_MEDICAL 3
+#define REPORT_ICON_RECON "\A3\Ui_F_Curator\Data\Logos\arma3_curator_eye_32_ca.paa"
+#define REPORT_PRIOITY_RECON 1
+
+switch (GVAR(state)) do {
```
AACO: While an integer switch makes logical sense here, since this is getting called multiple times per frame should this be in the if () exitWith {}; form? It's supposed to be minorly more performant

PR #162, opened 2017-07-28
Add script created objects to curator

PabstMirror:
Needed for #160


PabstMirror:
The thing is we don't want to just add everything (like players)
I don't think this is that bad, if we add other ways to add units it's just one function call that needs to be added.

PR #172, opened 2017-08-25
Allow VZ99 Mags to be selectable in VA

AACO:
Refer to title. Also cleans up some description stuff. I know proxy on it's own doesn't make 100% sense, but might as well leave it in there for MM discretion?


PabstMirror:
I tested this and it seemed to work,
but I remember having problems with `type` a while back, like shells not loading or not being simulated right

AACO:
I did testing locally/local dedicated and it seemed to load and fire correctly. I suppose there could be a networked issue but I have no idea how we'd test that...

PR #173, opened 2017-08-27
Reflow assign gear for JIPs

### LINKED ISSUE #94: Potato assignGear vs ArmA gear initialization race condition.

AACO:
Resolves #94 

This PR will:
- Move assignGearMan to initPost
- Add actual JIP handling (tested on local dedicated, should be interesting in large scale dedicated)
- Move google handling to assignGearMan
- Refactor the reset gear menu tool to use the new function
- Ensure assign gear is only run on the server to better leverage caching

If this works reliably in session, we should be able to remove the watch dog (might need to keep it for AI?)

Potential downsides: 
- If the server initPost actually works on JIPs (which is historically has not) assign gear could run twice per JIP (with no benefit)
- Moving the googles stuff to assign gear might lead to strange stuff for AI?

### DIFF
--- a/addons/adminMenu/XEH_postInit.sqf
+++ b/addons/adminMenu/XEH_postInit.sqf
@@ -14,14 +14,8 @@
 
     [{
         if (missionNamespace getVariable [QEGVAR(assignGear,usePotato), false]) then {
-            // set unit loadout overrides our sick shades :(
-            private _goggles = goggles _this;
-            removeGoggles _this;
-
-            diag_log text format ["[POTATO] Calling %1", QEFUNC(assignGear,assignGearMan)];
-            [_this] call EFUNC(assignGear,assignGearMan);
-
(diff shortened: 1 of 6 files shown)


PabstMirror:
I'm willing to give this a try,
gear has been so finicky about when it's run, it would be nice to have more confidance in our system and not need the fallback hack.

One of the problems I remember seeing before is extra backpacks on the ground,
and we should also see how this handles "pre-filled" backpacks.

AACO:
In testing it seemed okay for prefilled junk, however even with the current system we see AI backpack issues (wrong back pack, dropped back pack, etc) with low FPS/high player count.

PR #190, opened 2017-10-03
3den gear fix

### LINKED ISSUE #189: Assign gear no longer runs in 3den for new units
Either post 1.76 or post CBA update assign gear is no longer being run on newly created units in 3den.

AACO:
Resolves #189 
- Runs init for 3den
- Removes watch dog

### DIFF
--- a/addons/assignGear/functions/fnc_assignGearMan.sqf
+++ b/addons/assignGear/functions/fnc_assignGearMan.sqf
@@ -18,8 +18,8 @@
 
 TRACE_1("params",_this);
 
-params ["_unit", ["_request", false, [false]]];
-TRACE_3("",_unit, local _unit,_request);
+params ["_unit"];
+TRACE_2("",_unit, local _unit);
 
 BEGIN_COUNTER(assignGearMan);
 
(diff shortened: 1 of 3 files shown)


addons/assignGear/functions/fnc_assignGearMan.sqf
```diff
@@ -63,21 +63,3 @@ if (isText (_path >> "init")) then {
...
-
-    private _arrayUniform = (_loadoutArray select LA_UNIFORM_INDEX) param [0, ""];
-    private _arrayVest = (_loadoutArray select LA_VEST_INDEX) param [0, ""];
-    private _arrayBackpack = (_loadoutArray select LA_BACKPACK_INDEX) param [0, ""];
-    TRACE_6("Checking loadout: Uniform [%1-%2] Vest [%3-%4] Backpack [%5-%6]", _arrayUniform, uniform _unit, _arrayVest, vest _unit, _arrayBackpack, backpack _unit);
-
-    if ((_arrayUniform != (uniform _unit)) || {_arrayVest != (vest _unit)} || {_arrayBackpack != (backpack _unit)}) then {
-        ERROR_2("Mismatch [%1] [%2]", name _unit, typeOf _unit);
-        _unit setUnitLoadout _loadoutArray;
-        ["potato_adminMsg", [(format ["Auto-reset gear on %1", name _unit]), "Server"]] call CBA_fnc_globalEvent;
```
PabstMirror: I saw this run a few time last session, maybe we should keep it around until we're sure it's working

AACO: Most of the issues are long loadin (where it runs on the server, but doesn't actually set gear) the one corner case I normally see is DC on loadin so the first init runs, misses the client, and the client post init never runs, leaving a future dude dead at spawn

PR #192, opened 2017-10-23
AssignGear - Give side specific UAVs

PabstMirror:
Loadouts still have side specific uav types
```
class soldier_UAV_F: Soldier_F {
  backpack[] = {"O_UAV_01_backpack_F"};
  linkedItems[] += {"O_uavterminal"};
```


AACO:
The code looks good, it will need to be updated if/when we push additional blu/ind factions. However is this something we want to do? There are other 'faction/side specific items that won't be changed, i.e. vehicle keys and chemlights. Do we want automagic on these items, or should it be up to the mission maker on those items?

PabstMirror:
We could solve this via macros in loadouts
```
class blu_f { // BluFor
    #define ITEM_SIDE_PREFIX B
    #include "Loadouts\ru_ak74_floral.hpp"
    #undef ITEM_SIDE_PREFIX
};
```
Would that be better, or get too ugly?

PR #195, opened 2017-11-22
Standardize all weapon's AI max range

TheCandianVendingMachine:
Standardizes all weapons to have basic AI range modes. Custom python script included to make generation easier.

Why:
Currently, due to different mods and standards, there are many issues regarding what guns will shoot at what ranges (with the AI). This PR fixes all inconsistencies and allows for easy mass changes if we determine if a weapon is currently shooting too far/short. Example of weapons that have bad configs:

`rhsusf_weap_m1911a1::Single::maxRange = 600, 1/25 chance`. The AI has a 1/25 chance of shooting its pistol up to 600 meters (Reason why we sometimes see AI take its pistol out instead of its rifle, even when it has full ammo)

`rhs_weap_akm::Single::maxRange = 400, 3/10 chance -- CUP_arifle_AKM::Single::maxRange = 600, 1/25 chance `. Two of the exact same guns perform differently due to different ideologies of what the max range should be. This fix will make it so both weapons will have the exact same range, unless otherwise configured.

`rhs_weap_m4::Single::maxRange = 500, 1/5 chance -- CUP_arifle_M4_Base::Single::maxRange = 300, 1/20 chance` Exact same reason as before.

By default, all rifles and carbines will have the AI have the possibility to shoot up to 500 meters with iron-sights maximum, and can range up to 650 if they have magnified optics.

All LMG's and MMG's can range up to 800m with iron sights, with LMG's having less of a chance to do so than an MMG. This range stays the same, but has a better chance of hitting with magnified optics attached.

Every SMG can range out to 150 meters maximum

Every DMR  can range out to 1100 meters with magnified optics by default

Every sniper can range out to 1800 meters with magnified optics by default

Every pistol and machine pistol can range out to 80 meters by default

Every shotgun can range out to 120 meters by default

Why did I pick 500 meters engagement range for all rifles? I figured through empiracle evidence that in session we engage enemies up to 600 meters maximum, with anything less than that being within our engagement range. We regularly drop AI without them shooting back because of the limits imposed by modded and vanilla weapons. Since 600 meters is not the normal max range we kill them at, I decided that 500 meters would be a good compromise.

Similar reasons as for the LMG's and MMG's max range is 800 meters, but the reason they are punched out is because they are much less used than carbines and rifles on the AI. I also, personally, believe that LMG's should feel comfortable engaging anything less than 800 meters which factored into this decision.

cfgWeapons.hpp:
Every single weapon that has a custom mode defined inside is generated and put into this file. Each weapon then gets relevant modes added to it and defined. Each new mode inherits off of the first mode that each weapon has defined in its file. This can lead to issues, but I can't think of a better way to add sound to AI weapons than this.

weapons.txt
Only new file added. Inside defines each weapon with a defined mode, and categorizes them into unique sections. This file is only used for the generation script. The script works by reading the base config (bracket arrays defined after the `[weapon_type]`) and adds it to a weapon object, and if a weapon has a `= {bracket_array}` after it, the object will gain the data of that instead.

The other section is all of the custom modes being added. `InheritMode` is after the class definition as for this gets replaced by the first default mode the gun has. Each mode reads from the bracket array by its 0-indexed line position, starting from the first definition.

How to use populateCfgWeapons.py:
Launch the script with arguments of `[path_to_unpacked_pbos, path_to_cfg_weapons_output, output_file_name]`. `path_to_unpacked_pbos` is where all the unpacked PBO's are. The script reads through the directories, finds each `cfgWeapons.hpp` and then reads weapon data. The script requires unpacked pbo's as for I did not figure out how to use `unpackPbo` via command line.

Once run, the script will generate `cfgWeapons.hpp`. If any weapons are found, and not defined in `weapons.txt` the script will not generate the weapon and instead print a warning about the weapon, including its full classname for easy copy/paste into `weapons.txt`


TheCandianVendingMachine:
Another plus to this is that COOP's on Takistan and other desert maps will become more viable since the AI will shoot us back

AACO:
- The addon needs to be camel case, like every other addon (`weaponCfgFixes`)
- I feel like the txt configuration file is a pretty big weak point to this item
- The python script should be in the tools folder, not the addon folder
- The `requiredAddons` section is missing the addons required by the config
- Since it does require mods, needs to add`POTATO_LEAN_CONFIGS` defines (example: https://github.com/BourbonWarfare/POTATO/blob/master/addons/miscFixes/config.cpp)
- I have some minor concerns on redefining all these weapon's relations, because if A3/Mods change parents, we would need to update this file, otherwise we'd be updating the class's base class, which can break stuff.
- Does this handle the attached weapons, like UGLs or the underbarrel .50?
- If the modes are additive, does it actually fix the issue with the 600m pistol shooters?

Can you post some use case differences to show the effect this has on AI? I know the `aimingAccuracy`, `spotDistance`, and `spotTime` play into this as well.

I personally think the python script makes more sense as a SQF script, since you can easily check the weapon's base type so you can do base logic there (with limited overrides), so you don't have to try and maintain a config list.

@PabstMirror Do you think this would be better in miscFixes, or should it stay it's own addon?

TheCandianVendingMachine:
> I feel like the txt configuration file is a pretty big weak point to this item

I 100% agree, and I tried to find a way around it to avoid it. The only issue that existed is that I needed to figure out what type of weapon each one was so I can set the config accordingly, and there is nothing in the configs that would help me with that.


> Does this handle the attached weapons, like UGLs or the underbarrel .50?

It does not handle it; primarily because not all weapons define their underbarrel in their respective class, or it doesn't define it at all (looking at your RHS). It would require more definitions in the `weapons.txt` file, or if I switch to SQF a way to figure out if its a grenade launcher


> If the modes are additive, does it actually fix the issue with the 600m pistol shooters?

I forgot to mention in the PR that it parses through all the already defined modes, and sets the `aiMin/Mid/MaxRangeProbab` to zero to have the AI avoid using the default modes.

I will change the Python script to a SQF if you can find any unique way to figure out if a weapon is a LMG/MMG/Rifle/Etc. I wanted to go with the most dynamic way, and sadly the .txt and .py was the best way I could find.

AACO:
For partitioning weapons by type in SQF, you can use the same logic used in the A3 supply box storage editor ( https://pastebin.com/yCuf8pbx ). To get parents, you can use https://community.bistudio.com/wiki/inheritsFrom or a similar command.

Would it be better to `delete` the unused modes, instead of zero them out?

TheCandianVendingMachine:
>Would it be better to delete the unused modes, instead of zero them out?

No. I tried to delete them and replace them with custom ones, but the way that ARMA handle's weapon sounds is defined in the modes. Unless if we copy/paste the entire mode and just change the required values it would be not possible. The reason I don't just copy/paste the entire mode is to lower the line count in `CfgWeapons.hpp`. If you don't mind the huge (>20k) line count, I can modify it so we only use custom modes.

I will look at the pastebin that you linked, and will probably scrap the Python script in favour of SQF. How would you recommend when the generation happens, and where it outputs to? I would think at game-load time would be the best time to generate the data, but I don't know where it should output to. I don't think I can change Config values on the fly, right?

AACO:
The way some of the BI/ACE tools you just run through the debug console and it outputs to the clipboard.

TheCandianVendingMachine:
With the auto-generation script, I still need the definition of all classnames that will be generated. Otherwise there will be a 2.5Mb file with >100k lines. Necessary evil

Also, the CUP SCAR-L has a config issue that requires manual fixing to fix. There is also a small bug in which the RHS M4A1 will not get its sound in "Full Auto" mode unless if its manually fixed. Another small issue is that the base class `RifleCore` is defined last - needs manual fixing.

All in all, I think that this is satisfactory. If fixes most, if not all of the guns that we commonly use.

AACO:
`configProperties` https://community.bistudio.com/wiki/configProperties should allow you to ignore inherited values, so you should be able to leverage that for a smaller, more condense Cfg dump.

TheCandianVendingMachine:
With that latest commit the script is working as intended. It will now generate all applicable weapons, and can be expanded to include more definitions if wanted without a predefined list.

addons/weaponCfgFixes/PBOPREFIX
```diff
@@ -0,0 +1,2 @@
+z\potato\addons\weapon_cfg_fixes
```
AACO: This needs to be fixed for camel casing

addons/weaponCfgFixes/CfgWeapons.hpp
```diff
@@ -0,0 +1,14354 @@
...
+class CfgWeapons {
+    class Default;
+    class RifleCore: Default {};
+    class Rifle: RifleCore {};
+    class Rifle_Base_F: Rifle {};
+    class Rifle_Long_Base_F: Rifle_Base_F {};
+    class PistolCore: Default {};
+    class Pistol: PistolCore {};
+    class Pistol_Base_F: Pistol {};
+    class Rifle_Short_Base_F: Rifle_Base_F {};
```
AACO: Do these need full redefining, or can we do:
```SQF
    class Rifle_Base_F;
    class Rifle_Long_Base_F;
    class Pistol_Base_F;
    class Rifle_Short_Base_F;
```

I want to fuck with as few base classes as possilbe

addons/weaponCfgFixes/CfgWeapons.hpp
```diff
@@ -0,0 +1,14354 @@
+class Mode_FullAuto;
+class Mode_Burst;
+class Mode_SemiAuto;
+class CfgWeapons {
```
AACO: It might look strange in SQF, but I think we can cut this down a good bit by using defines

TheCandianVendingMachine: Can you elaborate on what you mean by "using defines"?

TheCandianVendingMachine:
To explain the issue that was occurring:
 If a weapon had a mode defined in its body, but not in its `mode[]` array, it would not be generated since why would any modder do that, it's non-nonsensical. It turns out some weapons (only `rhs_weap_m4_base`) defined a mode in its body, but not its array. `rhs_weap_m4a1` inherited the `fullAuto` mode defined in `rhs_weap_m4_base`, and because of that it didn't inherit its sounds because  it would be inherited from `Mode_SemiAuto`. The fix just generates the mode if it finds it by checking if a internal class has `sounds[]` defined inside, and if so it will generate the class body.

AACO:
A couple questions for clarity,

Why are we defining extra modes for weapons (we're defining the full set S/B/FA/MO/LO for every gun, even if only S/B/FA are defined). Does this actually help? How are optics defined for AI, will an AI with a 1x red dot get the long optic mode?

We're only extending the first mode, what happens if the sounds or other inherited values are different between the modes?

TheCandianVendingMachine:
So there are a lot of problems with generating the modes with an automated way. The only way around that is to manually update the list, with possibly a script to generate a template using the default values of the weapon modes.

What my original idea was with this was to have a script run through all of the defined modes, copy over the needed data, and then through manual intervention change any wanted values. This would get rid of the new mode definitions, the possible sound issues of inheriting from the first mode, and ensure that each mode is the same. However it comes at the cost of manual labour. It is easily posssible to reduce the labour to only having to define whatever new weapon is added, but it will still exist.

**Why can't we automate this:**
Every mod has different standards. Hell, ARMA 2 even had different standards that is reflected in CUP. For example, the CUP SCAR has it's modes defined as `SCAR_L_*Mode*` and `SCAR_H_*Mode*`. HLC has it's modes, in the G3, defined normally except for the occasional `AI_far` and `AI_Burst_Close`. Without maintaining a lookup table of each mode and what it corresponds to, we can't figure out what each mode means without possibly a very advanced neural network.

That's not even getting started on how we determine what each weapon is. The way `bis_fnc_itemType` works is by checking the defined value `cursor=` and if it equals an assault rifle, sniper, etc it will return that. This is a problem since, for example, the `rhs_weap_m107` is defined as a rifle and thus generates rifle data. We can't get around this without defining each weapon manually or maintaining a list of what each weapon type is.

Now this problem is more wide-spread than I originally thought. When trying to make a coop using only weapons that can reach out to 500 meters I found that none of the non-AK CUP weapons pass this. That means that a large majority of the weapons in our mod pack will have AI refusing to shoot past 400 meters, as demonstrated in...

**Video Demonstrating Values**

The current video I have to demonstrate the problem is flawed in the ranges claimed to be tested at, and uses calibre differences. The only use of the video is to demonstrate that with the changes the AI will engage back at ranges greater than almost all CUP weapons more accurately. The video has been made on our dedicated server WITHOUT the changes applied. I found a weapon that reflect the current status, and a weapon that has the values that I am changing everything to. There is also major audio-sync issues for the first three quarters of the video.

The video also demonstrates that AI with 5.56 weapons are lethal at the same range in later points in the video. The Blufor AI have RHS M16A4's that have properly configured values allowing them to engage up to, and past 500 meters.

The ranges are mostly accurate, but vary widely due to a lack of proper testing. I can re-do the video, but the current one demonstrates the differences in weapons just fine. If you are concerned about the weapon that the Indfor AI have, just imagine that instead of the CZ805 they have that instead they have a M16A2 or a CUP M4A1; the problems exist just the same for all of those weapons.

https://youtu.be/0VHEcOhlJaY

TheCandianVendingMachine:
Because I am a crazy person, I have went through and changed all cfgWeapon data to the new data manually. The script now only copies the config data to cfgWeapon's, and all relevant modes. I have also added one new mode, `potato_long` to all M/LMG's to expand their max range to 800 meters.

All new weapons will need to be modified manually if there is the thought that they have "bad" configs. The script does not generate new data.

I also looked into it further, and found that basically all RHS AFRF and GREF configs for weapons have their max ranges at 400 meters, and basically all CUP rifles have their max ranges limited to 400 meters. HLC weapons have data that allow all rifles to shoot up to 500 meters, and most RHS USAF weapons (M4, M16, and HK variants) have "good" configs.

If any new weapons get added or its class-name changed, and a maintainer does not want to update this file I will be happy to do so and make a new PR if this is adopted.

TheCandianVendingMachine:
Closing to combine with ammo fixes

PR #200, opened 2017-12-18
AI Cfg Fixes

### LINKED ISSUE #198: AI don't work the same way on the HC versus not on the HC
I was testing some missions and I found a very curious issue with AI reacting on the Headless Client versus them not on it. It seems that AI on the headless client will not engage back at players who shoot at them, while AI not on the headless client will shoot back in the same circumstances. AI on the HC however will shoot back if you have another HC AI shoot at them via the "Suppress Fire" module.

All tests have been conducted on the BW server with BWMF on all.

Videos of problems:
[Mission: brdn_ai_reveal_problem.Chernarus - No HC module placed](https://youtu.be/9Z4QwgJlDhM)
[Mission: brdn_ai_reveal_problem_2.Chernarus -  HC module placed](https://youtu.be/o69kmZQYQ8w)

Now in my testing I determined that this is probably a locality issue with HC units not getting data from non-HC units, however they will receive data the other way around as showcased in the last test of the second video. I can't test this on Vanilla due to me lacking a server with a headless client

I don't know how to fix this, if we can fix it, or if this is intentional - I just wanted to point out an issue that I discovered.

TheCandianVendingMachine:
Fixes weird configs in mods to make the AI perform better.

Resolves #198


addons/aiCfgFixes/functions/fnc_generateAmmo.sqf
```diff
@@ -0,0 +1,96 @@
...
+ * Example:
+ * [] call potato_weaponCfgFixes_fnc_generateAmmo;
+ *
+ * Public: Yes
+ */
+#include "script_component.hpp"
+#define LOWEST_BASE_CLASS -2
+
+#define CFG_WEAPONS_HEADER "class Mode_FullAuto;" + LINE_BREAK + "class Mode_Burst;" + LINE_BREAK + "class Mode_SemiAuto;" + LINE_BREAK + "class CfgWeapons {" + LINE_BREAK
+#define DEFAULT_MODE_VALUES INDENT + INDENT + INDENT + "minRangeProbab = 0; midRangeProbab = 0; maxRangeProbab = 0;" + LINE_BREAK
```
AACO: Are these defines relevant here?

addons/aiCfgFixes/CfgAmmo.hpp
```diff
@@ -0,0 +1,613 @@
...
+    };
+    class ShotDeployCore: Default {
+        audibleFire = 16;
+    };
+    class ShotgunCore: Default {
+        audibleFire = 16;
+    };
+    class ShellBase: ShellCore {
+        audibleFire = AI_AUDIBLE_FIRE_0;
+        GVAR(macroUsed) = "AI_AUDIBLE_FIRE_0";
```
AACO: Is there any non debug reason to have this value?

TheCandianVendingMachine: The Macro -> So if we find that the value I used is too extreme we can tweak all applicable values en-masse

`GVAR(macroUsed)` -> needed for the script to add the appropriate macro to the `avaliableFire` variable

addons/aiCfgFixes/PBOPREFIX
```diff
@@ -0,0 +1,2 @@
+z\potato\addons\aiCfgFixes
+
```
AACO: This file needs to be $PBOPREFIX$

addons/aiCfgFixes/XEH_PREP.hpp
```diff
@@ -0,0 +1,4 @@
+TRACE_1("",QUOTE(ADDON));
+
+PREP(generateWeapons);
+PREP(generateAmmo);
```
AACO: This file should have a newline at the end

PR #202, opened 2017-12-24
Use ace's getVehicleIcon

PabstMirror:
interaction menu will be using this, so everything should already be cached


AACO:
What wack job wrote that function?

PR #204, opened 2018-01-07
Add engine script setting

AACO:
Adds a setting that turns off RHS engine startup delay scripting. Discussion point: should we leave this defaulted to true, then update it in BWMF?

Also cleans up STUI variables that are unused.


PabstMirror:
I'd say just turn it off altogether, skip the setting

Alex6511:
Just turn it off this script sucks.

PR #205, opened 2018-01-07
Move to CBA system

AACO:
Instead of using the old ACE system (and hope it stays/remains unchanged) move to direct CBA settings calls.

Removed settings:
- ace_medical_bleedingCoefficient  (should be handled currently in BWMF)
- ace_medical_healHitPointAfterAdvBandage (should be handled currently in BWMF)
- ace_overheating_unJamOnreload (overheating is disabled)
- ace_overheating_unJamFailChance (overheating is disabled)

Changed settings:
- ace_frag_maxTrack 2 -> 5 (should this be closer to 10?)

Question: Are there forced settings below that should be done in BWMF?
Question 2: Should this be in post init?


PabstMirror:
Wrap in a `if (isServer) then {` or it will do a admin logged save-setting thing
I was thinking prioity should be 0, but I'm not sure,
also not sure when to run, I think pre-init is the best, but I think we'd still want the requiredAddons

AACO:
Updated, does DEBUG_SETTINGS do anything anymore?

PabstMirror:
i dunno, we can remove them

addons/settings/XEH_preInit.sqf
```diff
@@ -0,0 +1,38 @@
+#include "script_component.hpp"
+
+// turn off RHS engine startup script
+RHS_ENGINE_STARTUP_OFF = 1; // any non-nil value turns this off
+
+if (isServer) then {
+    // set STUI settings for the BW experience (occlusion set mission side)
+    ["sthud_settings_squadbar", false, 0, "server"] call CBA_settings_fnc_set;
+    ["sthud_settings_removedeadviaproximity", true, 0, "server"] call CBA_settings_fnc_set;
```
PabstMirror: As far as I can tell this setting has no effect on anything

AACO: Huh would you look at that... No only does it not do anything (from what I can see) the setting itself sets a slightly shorter GVAR (that also does nothing)....

AACO:
Just curious, what does moving it to its own includable file give us. Also if we're defining the settings array above, should we just append to the variable instead of overwrite?

PabstMirror:
If anyone else is using this it should make it easier to change a single `#include` line

AACO:
I agree with that convention for changing pre-existing settings, but if they blow away those settings the addon will not properly function.

It does give better readability convention if you want to view an addon's settings though.

PR #206, opened 2018-01-07
Move safe start to cba settings

AACO:
Does what it says on the tin, had to add a small backwards compatible snip to work with current BWMF (is there a better way pabo?)


PabstMirror:
See https://github.com/acemod/ACE3/blob/master/addons/markers/ACE_Settings.hpp

PR #208, opened 2018-01-07
Update recruit tags

AACO:
- Moves from ACE settings to CBA settings
- Splits draw 3d and the dialog logic/functionality
- Removes some math for the tag size
- Adds some math for the tag position


addons/recruits/XEH_preInit.sqf
```diff
@@ -6,32 +6,46 @@ PREP_RECOMPILE_START;
...
+        ["Show Recuit Message", "Show the recruit message dialog."],
+        "POTATO UI",
+        true,
+        0,
+        {
+            if (missionNamespace getVariable [QEGVAR(safeStart,startTime_PV), -1] != -1) exitWith {
+                [_this] call FUNC(toggleRecruitInfo);
+            };
+        }
+    ] call cba_settings_fnc_init;
```
PabstMirror: settings should be added on all machines

AACO: Do I have to worry about UI code running on server/HC?

PabstMirror: safer to run the setting init on the server as well

AACO: Yeah, I agree they need to be set everywhere, more concerned about running display code on headless

PR #210, opened 2018-01-07
Move briefing vars to 3DEN/GVARs

AACO:
It didn't seem to make sense to have these settings as CBA settings (besides the fact that it's easier). So this will change them to custom 3DEN settings based off the current 3DEN briefing menu.

Image of settings

It doesn't add too much to the overall editor experience, but it's better than burring it in a (now) very wide CBA settings menu.


addons/briefing/XEH_preInit.sqf
```diff
@@ -6,4 +6,7 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+GVAR(brief_addCredits) = false;
```
PabstMirror: I don't think this is safe for jips as it will overwrite the `setVariable [,,true]`

addons/briefing/XEH_postInit.sqf
```diff
@@ -11,12 +11,20 @@ if (hasInterface) then {
         [_unit] call FUNC(addIntentToZeus);
     }] call CBA_fnc_addEventHandler;
 
+    // TODO: BWC, remove this a while after (can't use movedToSQF = 1 here)
```
PabstMirror: I don't think we need this at all, beccause ace will auto conver the ace_settings from the description.ext to new cba settings
So we should just get BWC for free

AACO: I tried that with safe start and it was a no-go. I can give it another go, but not only did it not set the setting correctly, it overwrote the setting to junk.

addons/briefing/XEH_preInit.sqf
AACO: You're right, I'll rebuild the references with getVariable where if it's missing it defaults to false.

PR #213, opened 2018-02-06
Move to cba settings, refactor UI

AACO:
- Move to CBA settings from ACE settings
- Make the UI more full featured
- Add `SUPPRESSION` and `PATH` AI behaviors

TODO:
- Figure out BWC `movedToSQF` appears to be broken ATM

Question (@zerithsmash ):
- Are these default skills too aggressive? They're a decent bump.

Quick reel of the changed UI:
https://youtu.be/L8IDYu1cQ88


AACO:
Updated sliders:
https://i.imgur.com/MNWJaWK.png

addons/missionModules/ACE_Settings.hpp
```diff
@@ -1,128 +1,6 @@
 class ACE_Settings {
+    // TODO: BWC, remove this a while after
     class GVAR(aiSkill_set) {
-        category = QUOTE(PREFIX);
-        displayName = "Global AI Set Skill: Master Enable";
-        typeName = "BOOL";
-        value = 0;
-    };
-    class GVAR(aiSkill_aimingAccuracy_min) {
-        category = QUOTE(PREFIX);
```
PabstMirror: Would we want `movedToSQF` on all of these?
I think the mission defined ace_settings will cause problems without

AACO: Yeah I wanted CBA to overwrite the old mission values, but I would have had to change the variables. movedToSQF is needed for sure.

PR #214, opened 2018-02-09
Fix Burst-Value Oversight on CUP MG36, CUP XM8, and RHS M27 IAR

TheCandianVendingMachine:
Added script to check for this in future updates

Note: HLC apparently has a MG-36 in the class names, but it doesn't show for the user. The script shows the weapon as false positives. Not fixing for now just so when it gets added this is already compatible


AACO:
1. if those are pre-existing modes that users are using, we shouldn't be overriding the burst number at all (even if we're setting it to the current value)
2. HLC 100% has a user accessible MG-36. The MG-36 CMAG is the default HLC MG-36, but if you load a normal PMAG into the MG 36 the custom magazine scripting replaces your gun with the 'normal' MG-36 (inverse of the G-36 logic where the 'normal' gun is the default and is replaced by the CMAG version if the CMAG is loaded)

TheCandianVendingMachine:
1. A necessary evil. I can't filter out the difference between my generated modes and the mod creator modes. Unless if you want to manually scrape the file each time to get rid of `burst =` on the mod created values you need to print it.

2. I was basing my MG-36 knowledge off a previous version. I didn't realize that it was visible. Fixed with a new generated value

AACO:
1. Check `showToPlayer `? and don't touch the burst value (or any player affecting values) if it's 1?

TheCandianVendingMachine:
OK Changed it so if `showToPlayer == 1` I don't generate the burst value. This means that it will be whatever the mod creator wants it to be

PR #221, opened 2018-02-20
Add sunrise/sunset times to credits

PabstMirror:
I think this is information you would generally have a basic idea of


AACO:
I agree it's nice to have, but I don't know if it belongs in the credits. Maybe a a slightly more full featured weather forecast tab? Sunrise,  sunset, current conditions forecast conditions?

PabstMirror:
person living in a hut might not have weather forecast
but would still probably know sunset time is roughly same as yesterday

AACO:
I would argue that your average warfighter would have some level of access to forecasting, either way I don't think this content makes sense being in the credits tab.

Alex6511:
👍 for weather forecast tab, could probably just set it up to pull directly from the forecast settings in the mission too.

I don't check the weather every day but I can tell you that in Florida on August 1st it will be cloudy with a chance of showers.

AACO:
I guess it fits decently with map name information... I still think having a forecast tab would be worth while eventually, maybe improvement down the road.

PR #224, opened 2018-02-27
Normalize hc pos

AACO:
This is a very mean merge


AACO:
My only concern is off map/out of AO assets messing this up.

PabstMirror:
Could ignore aircraft
or run two passes and ignore groups far away from the first average
or set to the average of ai-average + player-average

But this is probably a good start,
could maybe switch to recursive calls to cba_fnc_waitUntil instead of pfeh

AACO:
Should be good for a final review @PabstMirror I added a module to force HC position if manual control is desired/required.

PR #226, opened 2018-04-10
Minor tweaks and fixes

AACO:
- Slightly buffs performance factor
- Majorly buffs recovery factor
- Some fixes/cleanup to the spectate unit init flow
- Minor tweaks to respawn (delay by each unit overall, instead of unit in group, increase delay, add delay to slotting to try and avoid one person slotting multiple times)
- Adds the importable base respawn configs


addons/spectate/functions/fnc_initSpectatorUnit.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Arguments:
+ * 0: Spectate unit <OBJECT>
+ *
+ * Examples:
+ * [player] call potato_spectate_fnc_initSpectatorUnit;
+ *
+ * Public: No
+ */
+
```
PabstMirror: `if (!hasInterface) exitWith {};`
might need this or dedi server might wait forever, not sure

AACO: Init should be local to the unit, but it's a good saftey check

AACO: Init is not local, so this is a very good check

AACO:
Tested the spectate/respawn changes, stamina should be in a lot better shape

PR #231, opened 2018-04-28
uniform>vest>backpack not vest>backpack>uniform

AACO:
I noticed this caused some issues with MMG/MAT magazines, since they only fit in backpacks, so try to fit smallest to largest

### DIFF
--- a/addons/assignGear/functions/fnc_getLoadoutFromConfig.sqf
+++ b/addons/assignGear/functions/fnc_getLoadoutFromConfig.sqf
@@ -72,14 +72,14 @@ _configItems = [_configItems, _containersArray, VEST_INDEX] call FUNC(addItemsTo
 _configItems = [_configItems, _containersArray, BACKPACK_INDEX] call FUNC(addItemsToContainer);
 TRACE_1("Remaining items: ", _configItems);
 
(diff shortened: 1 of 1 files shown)


addons/assignGear/functions/fnc_getLoadoutFromConfig.sqf
```diff
@@ -72,14 +72,14 @@ _configItems = [_configItems, _containersArray, VEST_INDEX] call FUNC(addItemsTo
 _configItems = [_configItems, _containersArray, BACKPACK_INDEX] call FUNC(addItemsToContainer);
 TRACE_1("Remaining items: ", _configItems);
 
+_configBackpackItems = [_configBackpackItems, _containersArray, UNIFORM_INDEX] call FUNC(addItemsToContainer);
```
PabstMirror: won't backpack medical go into uniform now instead of backpack?

AACO: Good call for backpack items

PabstMirror:
I still think uniform should be last
- it doesn't hold much so it shouldn't change much
- this is where radios will be removed and added via acre radio id assignment

PR #233, opened 2018-06-14
Make AI shoot rockets more often

TheCandianVendingMachine:
Currently AI will not fire MI-8 or MI-24 rocket pods at players if they are not in a vehicle. This changes that, and makes it so the AI will consider shooting regular RPG rockets at players as well

Why:
Currently when a MI-8 with rocket pods fly overhead players ignore it because they dont shoot their rockets. Make that assumption go away with this, and make COOPs more deadly

Custom ammo type implemented as to not break anything that uses R_80mm_HE


zerithsmash:
I'm ok with this if you are Pabst.

PR #236, opened 2018-08-31
AI Vehicle Bailing

TheCandianVendingMachine:
Makes it so the AI will bail out of mission destroyed vehicles

How is this different to allowCrewInImmobile? That runs when a single wheel is destroyed, and does not care if the vehicle cannot shoot. This implementation makes it so:
  -If the engine is fully destroyed/the driver is dead they have a 80% chance of bailing out, otherwise they will bunker
  -If the turret/gun is fully destroyed/the gunner is dead they have a 50% chance of bailing out, otherwise they will ROUT
  -If the turret/gun AND engine is destroyed they WILL bail out

This makes a much more immersive environment where you don't need to blow up the vehicle in order to disable it. Mission-Killed vehicles

Eliminates the need for the event handler to disable allowCrewInImmobile in BWMF

This applies to: All wheeled APC's (BTR, Vodnik, Stryker, but NOT UAZ, Humvee, etc) and all Tanks/IFV's

There is a setting allowing you to enable/disable this during session or as a mission setting

Do we want this?

Example of this in action
https://youtu.be/oGwD30Yhtg8


addons/aiVehicleBail/XEH_postInit.sqf
```diff
@@ -0,0 +1,31 @@
...
+    
+    unassignVehicle _crewman;
+    _crewman leaveVehicle _vehicle;
+    doGetOut _crewman;
+    
+    private _angle = floor (random 360);
+    private _dist = (30 + (random 10));
+    
+    private _xEscape = _cX + (_dist * cos _angle);
+    private _yEscape = _cY + (_dist * sin _angle);
```
AACO: https://community.bistudio.com/wiki/getPos alt version

addons/aiVehicleBail/XEH_postInit.sqf
```diff
@@ -0,0 +1,31 @@
+#include "script_component.hpp"
+
+[QGVAR(bailOut), {
+    params["_center", "_crewman", "_vehicle"];
+    
+    if !(isPlayer _crewman) exitWith {};
+    
+    private _cX = _center select 0;
+    private _cY = _center select 1;
```
AACO: https://community.bistudio.com/wiki/params

PR #240, opened 2019-03-27
Add Vehicle Cookoff that doesn't override destroyed state

TheCandianVendingMachine:
Adds cookoff to vehicles when they receive damage to certain spots. Because this doesn't modify the damage vehicles will not be forced to cook off in order to die.

Features:
  -Cook-off fire that destroys the vehicle after 15 seconds (occupants can bail out)
  -Cook-off detonations that leaves the vehicle "alive" (occupants dies instantly)
  -Both things can happen during the same cook-off sequence

The system works by rolling a dice every time certain spots are damaged, and if the check is high enough the vehicle will cook off. The dice roll is influenced by the ammo that hit it (sort of), how much ammo and how much fuel is in the vehicle depending on where it gets hit. Unfortunately this means that each vehicle has to be configured for this, but it won't break anything if it isn't.

Why do we need this? Because cooking off vehicles makes them more vulnerable to infantry and it looks pretty rad.

https://youtu.be/5RmyHer-cp4

### DIFF
--- a/addons/aiVehicleBail/functions/fnc_handleCookoff.sqf
+++ b/addons/aiVehicleBail/functions/fnc_handleCookoff.sqf
@@ -73,7 +73,7 @@ private _alreadyDetonating = _vehicle getVariable [QGVAR(detonating), false];
 private _alreadyCookingOff = _vehicle getVariable [QGVAR(cookingOff), false];
 
 if (!_alreadyDetonating && { _chanceOfDetonate > random 1 }) exitWith {
-    [_vehicle, _currentAmmoCount] call FUNC(detonate);
+    [_vehicle, _currentVehicleAmmo] call FUNC(detonate);
     diag_log text format["[POTATO] (cookoff) Detonating [%1] with a chance-to-detonate [%2] hit [%3]", _vehicle, _chanceOfDetonate, _hitPoint];
     _vehicle setVariable [QGVAR(detonating), true];
 };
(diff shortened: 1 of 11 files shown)


addons/aiVehicleBail/functions/fnc_handleCookoff.sqf
```diff
@@ -0,0 +1,97 @@
...
+        _chanceOfDetonate = ([_vehicleConfig >> QGVAR(engineDetonationProb), "number", 0] call CBA_fnc_getConfigEntry) * _incendiary * _currentFuel;
+        _chanceOfFire = ([_vehicleConfig >> QGVAR(engineFireProb), "number", 0] call CBA_fnc_getConfigEntry) * _incendiary * _currentFuel;
+    };
+};
+
+private _alreadyDetonating = _vehicle getVariable [QGVAR(detonating), false];
+private _alreadyCookingOff = _vehicle getVariable [QGVAR(cookingOff), false];
+
+if (!_alreadyDetonating && { _chanceOfDetonate > random 1 }) exitWith {
+    [_vehicle, _currentAmmoCount] call FUNC(detonate);
```
PabstMirror: `_currentAmmoCount` is undefined
probably should be `_currentVehicleAmmo`

addons/aiVehicleBail/functions/fnc_handleCookoff.sqf
```diff
@@ -0,0 +1,97 @@
...
+    diag_log text format["[POTATO] (cookoff) Detonating [%1] with a chance-to-detonate [%2] hit [%3]", _vehicle, _chanceOfDetonate, _hitPoint];
+    _vehicle setVariable [QGVAR(detonating), true];
+};
+
+if (!_alreadyCookingOff && { _chanceOfFire > random 1 }) exitWith {
+    [_vehicle] call ACEFUNC(cookoff,cookOff);
+    if (!_alreadyDetonating && { [_vehicleConfig >> QGVAR(detonationDuringFireProb), "number", 0] call CBA_fnc_getConfigEntry > random 1 }) then {
+        [_vehicle, _currentAmmoCount] call FUNC(detonate);
```
PabstMirror: same

PR #246, opened 2019-07-12
Temp medical patch (server only)

PabstMirror:


addons/miscFixes/XEH_serverPostInit.sqf
```diff
@@ -1,5 +1,78 @@
...
+                diag_log text format ["[POTATO-Temp Ace Fix] fixing pain blackscreen effect"];
+                if (hasInterface && {ace_medical_feedback_painEffectType == 0}) then {
+                    ace_medical_feedback_ppPain ppEffectAdjust [1, 1, 0, [1, 1, 1, 0], [1, 1, 1, 1], [0.33, 0.33, 0.33, 0], [0.59, 0.64, 0, 0, 0, 0, 4]];
+                    ace_medical_feedback_ppPain ppEffectCommit 0;
+                };
+            }, [], 1] call CBA_fnc_waitAndExecute;
+        } remoteExecCall ["bis_fnc_call", 0, true];
+
+
+        private _sideCounts = [west, east, resistance] apply {
```
TheCandianVendingMachine: This seems really unreliable. If a mission has two playable sides in a coop (not common, but often enough to be an issue) it will trigger. If this is only temporary its fine, but maybe in the future there should be a function or a GVAR to determine if its a TvT from if there is a ratio or not.

PR #247, opened 2020-03-07
changes ready for ACE Vehicle Damage

TheCandianVendingMachine:
to be merged if/when ACE vehicle damage gets merged


TheCandianVendingMachine:
outdated. may fix later

PR #249, opened 2020-03-26
Give zeus access to server perf stats 

PabstMirror:
Added as tooltip to AI count display (will show HC when avail)
image

Do we still need the ACRE spectator button?
I think there is a ace itneract menu that does almost the same thing? 
but it might not hurt to keep the colored button to indicate the status


TheCandianVendingMachine:
I know I personally like Zeusing not in ACRE Spectator so I'd say keep the button there

PR #250, opened 2020-04-03
Experimental - Show reminder when non-crew uses vehicle they might not be trained on

PabstMirror:
Just a little reminder in the heat of battle (no hard lockout or anything like that) when regular soldiers hop in
- driver/pilot of aircraft or tracked vehicle
- gunner seat with a weapon that has big hit (missiles, cannon)
This isn't perfect, Bubba can probably figure out a SPG launcher, but might not have any idea how to use a 7.62 RCWS

Not sure if we actually want this, just an idea

### DIFF
--- a/addons/miscFixes/XEH_crewTraining.sqf
+++ b/addons/miscFixes/XEH_crewTraining.sqf
@@ -1,0 +1,59 @@
+#include "script_component.hpp"
+
+// PabstMirror - Show reminder when non-crew enters vehicles seats that they might not be trained on
+LOG("Adding Crew Training Setting");
+
+[
+    QGVAR(crewTraining),
+    "CHECKBOX",
+    ["Show Crew Training Reminder", "Show reminder about crew training when entering a complex vehicle weapon"],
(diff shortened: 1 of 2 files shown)


addons/miscFixes/XEH_crewTraining.sqf
```diff
@@ -0,0 +1,59 @@
...
+        if (!_enabled) exitWith {};
+
+        GVAR(crewTraining_peID) = ["turret", {
+            params ["_unit", "_turret"];
+            TRACE_2("turret EH",_unit,_turret);
+
+            if ((!alive _unit) || {_turret isEqualTo []}) exitWith { TRACE_1("dead or dismounted/cargo seat",_turret); };
+
+            private _unitType = toLower typeOf _unit;
+            if ((["pilot", "crew", "msv_cc", "msv_vic"] findIf {_x in _unitType}) > -1) exitWith { TRACE_1("trained",_unitType); };
```
Alex6511, line 29: I want to say there's a third MSV for vehicles by the way, but maybe you've reduced slots since I looked at this repo.

PR #251, opened 2020-04-12
Recruit Tracker to Discord Webhook - (dedicated server only) 

PabstMirror:
image

### DIFF
--- a/extensions/Webhook.cpp
+++ b/extensions/Webhook.cpp
@@ -7,7 +7,7 @@
 #include <future>
 #include "cpr/cpr.h"
 
-constexpr auto VERSION_STR = "v1.0.0";
+constexpr auto VERSION_STR = "v1.0.1";
 
 
 extern "C" {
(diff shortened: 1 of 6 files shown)


extensions/Webhook.cpp
```diff
@@ -0,0 +1,53 @@
...
+#define VERSION_STR "v1.0.0"
+
+
+extern "C" {
+    __declspec(dllexport) void __stdcall RVExtension(char* output, int outputSize, const char* function);
+    __declspec(dllexport) void __stdcall RVExtensionVersion(char* output, int outputSize);
+    __declspec (dllexport) void __stdcall RVExtensionRegisterCallback(int(*callbackProc)(char const* name, char const* function, char const* data));
+}
+
+int(*callbackPtr)(char const* name, char const* function, char const* data) = nullptr;
```
TheCandianVendingMachine, line 19: couldn't `std::function<int(char const*, char const*, char const*)> callBackPtr = []() -> int { return 0; }` work instead of a function pointer?

PR #253, opened 2020-09-01
"port" hit effects from ACE 2 into POTATO

TheCandianVendingMachine:
title. makes it easier to see your hits at distance


PabstMirror (approved):
looks neat, lets give it a try

PR #266, opened 2020-09-23
BWMF Rework - Potato Compatibility

Alablm:
Makes several changes across potato in order to support the BWMF Rework, see https://github.com/BourbonWarfare/bwmf/pull/149

Note: This will break old coops. For the most part TvTs should be unaffected.


Alablm:
This should not be merged until https://github.com/BourbonWarfare/bwmf/pull/149 is ready.

PabstMirror:
ref https://github.com/Alablm/POTATO/pull/4
I think this should allow semi-complete backwards compat with old missions
- Keep old `potato_msv_` for old coops
- translate faction/loadout names to old style to allow respawn/zeus spawning of units to use the old loadouts (only if new ones don't exist)
- minor syntax stuff

Alablm:
Yeah, tracking that, @TheCandianVendingMachine has a change (https://github.com/Alablm/POTATO/pull/3) that I'm probably going to implement in conjunction with this then send all of that back over to the main repo

PabstMirror (approved):

PR #269, opened 2020-09-30
Switch cswCompat to CUP's native version

PabstMirror:
CUP v16 has native csw compats based on ours


PabstMirror:
I want to get this merged before we finialize the bwmf stuff
shoud be bwc

PR #273, opened 2020-10-23
Cleanup side/systemChat

PabstMirror:
Reduce messages shown in chat (less distractions, better stream/recordings)
- Prevents accident use of side/global text chat
- Players can still send messages with force prefix (e.g. `force blu ready`)
- Reduces some "Player connecting" spam (will be improved in 2.02)


PabstMirror:
todo: 
- [ ] keep enabled for admins

Alablm:
This change blocks chat in the map screen, not sure if there's a way to differentiate between the map and mission proper, but if there is we should add in an if statement to allow chat on the map.

BadWolf64:
Could add a check for CBA_missiontime to fix the above issue. If less than or equal to 0 should allow for chat to still be enabled in map screen at briefing time.

kil0byt3:
is this still req? - feel like we rarely have issues with people using side

main part I am still interested in is the player connecting spam fixes

Alablm:
> is this still req? - feel like we rarely have issues with people using side
> 
> main part I am still interested in is the player connecting spam fixes

Its never been required, just nice to have. Throw in the eh check and it should be good to ship

PabstMirror:
changed it to only block `player x connecting` msgs
because some mods/missions may use systemChat for important things

PabstMirror:
test if need to also block "using modified data"

kil0byt3:
you want this merged?

PR #280, opened 2020-12-15
Naloxone

Alablm:
Adds Naloxone auto injectors to be used as a counter to Morphine. Vanilla ACE does not provide a medication that counteracts the effects of morphine, due to morphine's extremely long time in system, as well as its unique effects on blood viscosity, making it the only medication to affect blood pressure. Naloxone in the real world is a medication used to rapidly treat and reverse the effects of opioid overdoses. In this implementation, I've written it specifically to counteract the effects of ACE's implementation of morphine. This will have two primary use cases:
Giving medics treating an otherwise stable patient an option to increase blood pressure in a patient who had previously received morphine
Allowing rectification of mistakes made without requiring admin intervention (i.e. multiple morphine injections preventing a patient from waking up from the uncon state)


Alablm:
Not sure what the best way to handle the file structure in this was, felt wasteful to split everything up as its only one item, but if it's necessary I'll do it.

Alablm:
Retracting this, working on a better implementation.

PR #282, opened 2020-12-17
Potato groups

Alablm:
Adds groups of potato units to the groups tab in 3den and Zeus


addons/units/CfgGroups.hpp
```diff
@@ -0,0 +1,15 @@
+class CfgGroups {
+  #include "groups\groupDef.hpp"
+  #define CLASS_MACRO(unit) USMC(unit)
+  GROUP_DEF(WEST,"POTATO: West",w,1)
+
+  #undef CLASS_MACRO
+  #undef FACTION_LETTER
```
PabstMirror, line 7: I think these can be removed (pboProject complains)
```suggestion
```

addons/units/CfgGroups.hpp
```diff
@@ -0,0 +1,15 @@
...
+  #define CLASS_MACRO(unit) MSV(unit)
+  GROUP_DEF(East,"POTATO: East",e,0)
+  
+  #undef CLASS_MACRO
+  #undef FACTION_LETTER
```
PabstMirror, line 12:
```suggestion
```

PR #285, opened 2021-01-14
Adds toolkits to all vehicles

Alablm:
Mostly for pilots, but also just a nice QOL change for vehicle crews.


PabstMirror:
Added a `if (GVAR(alwaysAddToolkits)) then {` check so mission makers could disable if they wanted to for some reason.
Just add `alwaysAddToolkits = 0;` to CfgLoadouts (it defaults to true if not defined)

PabstMirror (approved):

PR #287, opened 2021-02-10
Create zeusHC modules from pre-placed vehicles

PabstMirror:
Adds zeusHC modules for mission vehicles at runtime.
Vehicles also show up in factory

Guide for Mission makers:
- Place vic with the camo you want
- Place a unit of the side you want near the vic
it uses the side of the crew or nearest unit to the vic to determine what side it is
once all zeuses have loaded in you can then delete them or do whatever

Vehicles should show up in HC spawn modules and in factory


PabstMirror:
Example

PR #293, opened 2021-03-17
AI Factory Updates

TheCandianVendingMachine:
* Change Paradrop script to be better; implements my paradrop script. No more helicopters hanging in the air awkwardly
* AI will now always occupy vehicle gunner seats while before they only spawned in IFVs. This should allow AI helicopters to have door gunners and AI cars to have gunners so they can defend themselves
* AI will now have option to use LAMBS waypoints to complete their mission
* Add check-box to allow Zeus to tell AI to execute order at target mark after dismount, or dismount location
* Add check-box to allow Zeus to tell AI to use LAMBS AI waypoints after dismount

These should be good changes that allow for better flow in missions. Nothing tested in multiplayer/dedicated server, I don't see why anything wouldn't work however. This should make AI a bit more realistic to fight against when using POTATO Factory

### DIFF
--- a/addons/zeusFactory/functions/fnc_ai_attack.sqf
+++ b/addons/zeusFactory/functions/fnc_ai_attack.sqf
@@ -7,7 +7,7 @@ private _radius = _group getVariable [QGVAR(radius), 200];
 private _attackTarget = _group getVariable [QGVAR(attackTarget), true];
 private _useLAMBS = _group getVariable [QGVAR(useLAMBS), true];
 
-TRACE_5("ai_attack",_leader,_group,_radius,_attackTarget,_attackTarget,_useLAMBS);
+TRACE_5("ai_attack",_leader,_group,_radius,_attackTarget,_useLAMBS);
 if (!local _leader) exitWith { WARNING_1("Waypoint script ran on non-local unit [%1]",_leader); };
 
 if !(_attackTarget) then {
--- a/addons/zeusFactory/functions/fnc_ai_attack.sqf
+++ b/addons/zeusFactory/functions/fnc_ai_attack.sqf
@@ -4,13 +4,19 @@ params ["_leader"];
 
 private _group = group _leader;
 private _radius = _group getVariable [QGVAR(radius), 200];
+private _attackTarget = _group getVariable [QGVAR(attackTarget), true];
 
-TRACE_3("ai_attack",_leader,_group,_radius);
+TRACE_4("ai_attack",_leader,_group,_radius,_attackTarget);
 if (!local _leader) exitWith { WARNING_1("Waypoint script ran on non-local unit [%1]",_leader); };
 
-private _currentBeacon = (entities QGVAR(attackBeacon)) param [0, objNull];
-if (isNull _currentBeacon) exitWith {
-    WARNING_1("No Becacons [%1]",_currentBeacon);
-};
+if !(_attackTarget) then {
+    // if we don't want to attack a target, search the area instead.
+    this call FUNC(ai_search);
+} else {
+    private _currentBeacon = (entities QGVAR(attackBeacon)) param [0, objNull];
+    if (isNull _currentBeacon) exitWith {
+        WARNING_1("No Beacons [%1]",_currentBeacon);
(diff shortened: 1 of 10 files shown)


BadWolf64:
These changes look really interesting. I should be able to give them a quick test tonight. Though honestly the changes are pretty straight forward. Good job.

addons/zeusFactory/functions/fnc_ai_attack.sqf
```diff
@@ -4,13 +4,24 @@ params ["_leader"];
 
 private _group = group _leader;
 private _radius = _group getVariable [QGVAR(radius), 200];
+private _attackTarget = _group getVariable [QGVAR(attackTarget), true];
+private _useLAMBS = _group getVariable [QGVAR(useLAMBS), true];
 
-TRACE_3("ai_attack",_leader,_group,_radius);
+TRACE_5("ai_attack",_leader,_group,_radius,_attackTarget,_attackTarget,_useLAMBS);
```
PabstMirror, line 10:
```suggestion
TRACE_6("ai_attack",_leader,_group,_radius,_attackTarget,_attackTarget,_useLAMBS);
```

addons/zeusFactory/functions/fnc_createTransport.sqf
```diff
@@ -4,8 +4,8 @@
 params ["_factoryLogic", "_transportType", "_group", "_side"];
 TRACE_4("createTransport",_factoryLogic,_transportType,_group,_side);
 
-([_factoryLogic,_transportType,_side] call FUNC(getTransportType)) params ["_vehType", "_crewType", "_maxCargoRoom", "_addGunner"];
-TRACE_4("getTransportType",_vehType,_crewType,_maxCargoRoom,_addGunner);
+([_factoryLogic,_transportType,_side] call FUNC(getTransportType)) params ["_vehType", "_crewType"];
+TRACE_4("getTransportType",_vehType,_crewType);
```
PabstMirror, line 8:
```suggestion
TRACE_2("getTransportType",_vehType,_crewType);
```

addons/zeusFactory/functions/fnc_createTransport.sqf
```diff
@@ -30,11 +30,28 @@ private _driver = _group createUnit [_crewType, (getPos _vehicle), [], 0, "NONE"
...
+
+{
+    private _config = configOf _vehicle;
+    _config = [_config, _x] call ACEFUNC(common,getTurretConfigPath);
+
+    if (((getNumber (_config >> "isCopilot")) == 0) && {count getArray (_config >> "weapons") > 0}) then {
+        _gunnerTurrets pushBack _x;
+    };
+    false
+} count _turrets;
```
PabstMirror, line 47: just fyi we've found that `...; false} count [` is actually slower than `forEach` now (it used to be faster)
I think there are some edge cases where it can be faster if you know for sure you have code that don't need the true/false at the end but it's still probably not worth it

addons/zeusFactory/functions/fnc_getTransportType.sqf
```diff
@@ -21,16 +21,5 @@ case (east): {"potato_e_" + _crewType};
...
-    _maxCargoRoom = ([_vehType, true] call BIS_fnc_crewCount) - 1;
-    if (_transportType in [TRANSPORT_APC_RTB, TRANSPORT_APC_FOLLOW]) then {
-        _addGunner = true;
-        _maxCargoRoom = _maxCargoRoom - 1;
-    };
-};
-
 // Return:
-[_vehType, _crewType, _maxCargoRoom, _addGunner]
+[_vehType, _crewType]
```
PabstMirror, line 25: 3rd arg of function still used here
`([_factoryLogic,_transportType,_side] call FUNC(getTransportType)) params ["", "", "_maxCargoRoom"];`
https://github.com/BourbonWarfare/POTATO/blob/7d37de240765fea41bf31f2bda6bab3b67d07a2e/addons/zeusFactory/functions/fnc_createGroup.sqf#L10

PabstMirror (approved):

PR #295, opened 2021-04-10
SafeStart additions

BadWolf64:
Adds a mission attribute to the "Mission Testing Attributes" for SS length and Mission Length.

Provides a warning to admins and Curators at 15mins before the end of the mission length, and one at the expiry of mission time. 

Creates a marker with the listed mission end time. 

Includes a fix for the forum link in the Mission Testers checklist. 

Known issues:
Will create two warnings if for whatever reason safestart is re-added and removed. Will not effect the time that the message is displayed. 
If in the case above, this does not update the marker. Have tried to get this working but for whatever reason it will not at this time. Might need some help. 

EDIT: https://github.com/BourbonWarfare/POTATO/pull/295/commits/9e1caa0f287c5b9b1b7fd2d663a38a66e945e722 With this commit I have added a check to `missionTest` that will give feedback if Mission Type is not selected and if Mission Length has not been entered.

### DIFF
--- a/addons/safeStart/functions/fnc_missionTimeWarning.sqf
+++ b/addons/safeStart/functions/fnc_missionTimeWarning.sqf
@@ -7,8 +7,7 @@
  * 0: Start mission length warnings <BOOL>
  *
  * Examples:
- * [true] call potato_safeStart_fnc_missionTimeWarning;
- * [false] call potato_safeStart_fnc_missionTimeWarning;
+ * [] call potato_safeStart_fnc_missionTimeWarning;
  *
  * Public: Yes
  */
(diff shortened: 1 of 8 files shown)


addons/safeStart/functions/fnc_missionTimeWarning.sqf
```diff
@@ -0,0 +1,74 @@
...
+    },
+    {
+        [
+            "potato_adminMsg",
+            [
+                "15 mins to mission end",
+                "Time Keeper"
+            ]
+        ] call CBA_fnc_globalEvent;
+        ["Time Keeper", "15 mins to mission end", 30] call BIS_fnc_curatorHint;
```
PabstMirror, line 41: I'm pretty sure `BIS_fnc_curatorHint` only has local effects, need to send via event to zeuses

addons/safeStart/functions/fnc_missionTimeWarning.sqf
```diff
@@ -0,0 +1,74 @@
...
+                        "Time Keeper"
+                    ]
+                ] call CBA_fnc_globalEvent;
+            ["Time Keeper", "Mission Time has expired", 30] call BIS_fnc_curatorHint;
+            },
+            [_missionStartTime,_missionLength]
+        ] call CBA_fnc_waitUntilAndExecute;
+    },
+    []
+] call CBA_fnc_waitUntilAndExecute;
```
PabstMirror, line 61: it's fairly minor, but I'd rather these be `CBA_fnc_waitAndExecute` since the time delay is static
I'm happy to make the switch

BadWolf64: Sure thing. Can't remember if there is a specific reason that I did it this way, but if you are happy to then please do.

BadWolf64: Looking back there is a reason that this was used and that was in the case that SS was turned off multiple times. If we use `CBA_fnc_waitAndExcute` the warnings will be displayed at multiple times which could be confusing. I don't mind using it but it could lead to confusion as I am not aware of a way of removing the previous `waitAndExecute` in the case that SS is turned back on.

BadWolf64: I have changed one of these to be `waitAndExecute` and added a check in the first `waitUntilAndExcute` so that if SS gets turned on it will not duplicate the commands. I don't know of a better way to get this done.

PabstMirror (approved):

PR #296, opened 2021-04-23
update AI cfg tweaks configs

TheCandianVendingMachine:
this is still needed because BI don't want to fix it

https://feedback.bistudio.com/T121635

fixes any and all issue with mismatched configs

### DIFF
--- a/addons/aiCfgFixes/script_component.hpp
+++ b/addons/aiCfgFixes/script_component.hpp
@@ -2,7 +2,7 @@
 #include "\z\potato\addons\core\script_mod.hpp"
 
 // #define DEBUG_MODE_FULL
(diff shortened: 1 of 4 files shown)


PabstMirror (approved):

addons/aiCfgFixes/script_component.hpp
```diff
@@ -2,7 +2,7 @@
 #include "\z\potato\addons\core\script_mod.hpp"
 
 // #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
+ #define DISABLE_COMPILE_CACHE
```
PabstMirror, line 5:
```suggestion
// #define DISABLE_COMPILE_CACHE
```

PR #301, opened 2021-05-02
Experimental: Potato loadouts

PabstMirror:
see example: https://github.com/BourbonWarfare/POTATO/tree/potatoLoadouts/addons/loadouts

don't worry, not meant to totally replace bwmf custom loadouts!
but this provides a way to host loadouts w/ photos
can download loadouts and drop into your mission,

OR

you can setup a static link to the loadout
```
  class potato_w { // BluFor
    using = "aaf_f2000_digi_w";
  };
```
then any updates made to the loadout would apply to the mission in the future
we could guarantee that we will never remove a loadout, 
so if we were to drop addons we could edit the static loadout so that it's still valid


Alablm:
So to clarify, the `using=` "command" would differentiate a potato based loadout from a mission based loadout yeah?

Would there be a way to implement this as a backup system? I.e. primarily use x custom loadout, if it fails fallback to y default (potato) loadout.

PabstMirror:
something like this?
if (anything in the loadout doesn't exist in current modset) then use `backup = "AAF..."`

Alablm:
Yeah exactly. If it's as simple as adding a line to CfgLoadouts then it theoretically ensures indefinite compatibility for most missions, even if you lose the intended factions.

PabstMirror:
maybe some ideas in here that we could use, but probably not going to be used as is

PR #304, opened 2021-05-22
Add overpressure deaths from explosives

TheCandianVendingMachine:
Calculates pressure at a given distance from the explosive and checks if it could be fatal. Real data used, so this _is_ realistic; see papers referenced in code

### DIFF
--- a/addons/miscMedical/XEH_postInit.sqf
+++ b/addons/miscMedical/XEH_postInit.sqf
@@ -198,12 +198,12 @@ if (isServer) then {
 
         private _objects = (ASLtoATL _lastPos) nearEntities [["CAManBase"], 50];
         {
-            [QGVAR(explosion), [_lastPos, _mass, _filler, _x], _x] call CBA_fnc_globalEvent;
+            [QGVAR(explosion), [_lastPos, _mass, _filler, _x], _x] call CBA_fnc_targetEvent;
         } forEach _objects;
 
         TRACE_4("explosion server",_explosive,_mass,_filler,count _objects);
 
-        [_lastPos, _mass, _filler] call DFUNC(generateVisuals);
+        //[_lastPos, _mass, _filler] call DFUNC(generateVisuals);
     }] call CBA_fnc_addEventHandler;
 };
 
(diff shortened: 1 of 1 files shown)


addons/miscMedical/XEH_postInit.sqf
```diff
@@ -71,11 +71,378 @@ if (hasInterface) then {
...
+        };
+
+        if (_mass <= 0) then {
+            TRACE_3("no mass",_mass,_filler,_explosive);
+            _mass = 1;
+        };
+
+        _mass = [_mass, _lastPos] call FUNC(calculateHemisphericalBlastWeight);
+
+        private _objects = (ASLtoATL _lastPos) nearEntities [["CAManBase"], 50];
```
PabstMirror, line 405:
```suggestion
        private _objects = (ASLtoAGL _lastPos) nearEntities [["CAManBase"], 50];
```

addons/miscMedical/XEH_postInit.sqf
```diff
@@ -71,11 +71,378 @@ if (hasInterface) then {
...
+        _origin = _origin vectorAdd [0, 0, 0.05];
+        // if we can see the unit, that means that there is always going to be a pressure wave
+        // if there is a pressure wave, use the closest distance regardless of whether or not it can be seen
+        private _canSee = false;
+        private _distance = 1e10;
+        {
+            if !(_canSee) then {
+                _canSee = ([] isEqualTo lineIntersectsObjs [_origin, _x, objNull, _unit, false, 16 + 32]) && { !terrainIntersectASL [_origin, _x] };
+            };
+            _distance = (_x distance _origin) min _distance;
```
PabstMirror, line 429:
```suggestion
            _distance = (_x vectorDistance _origin) min _distance;
```

addons/miscMedical/XEH_postInit.sqf
```diff
@@ -71,11 +71,378 @@ if (hasInterface) then {
...
+                    format ["Distance: %1m Pressure: %2Pa Fatality Chance: %3 PsI: %4", _distance, _pressureAtPosition, _chanceOfFatality, _pressureAtPosition / 6895]
+                ];
+            };
+        } forEach _blastData;
+    }, 0, [_position, _blastData, CBA_missionTime + 10]] call CBA_fnc_addPerFrameHandler
+};
+
+if (isServer) then {
+    // add explosion event handler
+    [QACEGVAR(frag,frag_eh), {
```
PabstMirror, line 387: I think there might be some thermobaric rounds that we'd want to over-pressure
but are set to not ace_frag, we might be able to tweak their configs to always frag track, but not create any elements

we can deal with this later

PabstMirror (approved):

PR #312, opened 2021-07-26
Mission setting to control rain and map sound levels

PabstMirror:


kil0byt3:
is the ACE Volume hotkey doing the same as this, maybe this is something that could be in ACE Volume instead

PR #314, opened 2021-08-17
Potato ammo

TheCandianVendingMachine:
Ammo is not balanced, but all magazine classnames are there

Every non-RHS weapon in our modpack is covered by this. It's trivial to swap to balanced ammo when the time comes


addons/massCompat/script_component.hpp
```diff
@@ -10,3 +10,220 @@
...
+MAGAZINE_WELL_TRACER_ALT_TYPE(ammoType,bulletCount,White,box,alt), \
+MAGAZINE_WELL_TRACER_ALT_TYPE(ammoType,bulletCount,IR,box,alt), \
+MAGAZINE_WELL_RELOAD_TRACER_ALT_TYPE(ammoType,bulletCount,Red,box,alt), \
+MAGAZINE_WELL_RELOAD_TRACER_ALT_TYPE(ammoType,bulletCount,Green,box,alt), \
+MAGAZINE_WELL_RELOAD_TRACER_ALT_TYPE(ammoType,bulletCount,Yellow,box,alt), \
+MAGAZINE_WELL_RELOAD_TRACER_ALT_TYPE(ammoType,bulletCount,White,box,alt), \
+MAGAZINE_WELL_RELOAD_TRACER_ALT_TYPE(ammoType,bulletCount,IR,box,alt)
+
+#define TRACER_CLASSES(baseClass)\
+class AMMO_CLASS(SCORE_3(baseClass,tracer,Red)): AMMO_CLASS(baseClass) { \
```
PabstMirror, line 198: We could add `displaynameshort="Tracer";` to these so it shows in the weapon display box, I think many vanilla mags do this.

TheCandianVendingMachine: Is it the bullet or magazine that displays it?

PabstMirror: must be on the mag, because the ammo isn't changed

PabstMirror (approved):

PR #318, opened 2021-09-10
Fix old mags

PabstMirror:
Fix m2000/zafir


PabstMirror:
we'll probably want to `scope=1` the odd `150Rnd_762x54_Box_Tracer_Red` from potato_misc eventually

PR #321, opened 2021-10-05
Drop RHS US and Gref

PabstMirror:
@TheCandianVendingMachine - I ran the sqf for cfgAiFixes, but not sure if it's ok


TheCandianVendingMachine:
There are still American RHS configs in CfgAmmo for some reason. It doesn't look breaking

PR #342, opened 2021-12-21
CWR Implementation

TheCandianVendingMachine:
VD, AI Cfg Tweaks, and POTATO Ammo magwells. All armoured vehicles excluding the Walker Bulldog inherit from CUP so my job was dead easy.


addons/cwrFixes/script_component.hpp
```diff
@@ -0,0 +1,8 @@
+#define COMPONENT jsrsFixes
```
PabstMirror, line 1:
```suggestion
#define COMPONENT cwrFixes
```

PR #364, opened 2022-05-07
Mission testing in 3Den & Time Keeper changes

DanThomas64:
Have added the ability to open the mission making checklist in 3Den. Available through [POTATO --> Open Testing Menu]

 - Time Keeper has been changed so that we no longer get weird values in the Attributes Menu. The defaults have been set to 0 in the attributes menu.
 - This means that they will also report in missionTest module if a value has been set. 
 - The time keeper now creates local markers so that they it can be moved around the map to make it easier to reference. 
       - I am a little concerned about the method used here. I don't believe that it will create a boat load of new markers for all players in the same spot, but it needs testing.
- Briefing is not longer Brieifing (Thanks kilo).
- Mission timers are also now in the MM report and will show in the testing menu for quicker reference. 
- The time keeper has been given some default values in the case that the mission length attribute is not set correctly by the MM. 35 mins for a TvT and 65 mins for a COOP.

### DIFF
--- a/addons/safeStart/functions/fnc_missionTimeWarning.sqf
+++ b/addons/safeStart/functions/fnc_missionTimeWarning.sqf
@@ -20,6 +20,7 @@ if !(isServer) exitWith {};
 params [];
 
 private _missionLength = getMissionConfigValue [QEGVAR(missionTesting,missionTimeLength),0];
+if (_missionLength isEqualType "") then { _missionLength = parseNumber _missionLength; }; // BWC with old attribute that was string
 private _missionLengthSec = _missionLength * 60;
 
 if (_missionLength == 0) then {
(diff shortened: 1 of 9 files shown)


addons/safeStart/functions/fnc_missionTimeWarning.sqf
```diff
@@ -19,15 +19,26 @@ if !(isServer) exitWith {};
 
 params [];
 
+private _missionLength = getMissionConfigValue [QEGVAR(missionTesting,missionTimeLength),0];
+private _missionLengthSec = _missionLength * 60;
```
PabstMirror, line 23: on old missions it's getting a string
`Error *: Type String, expected Number,Not a Number`

PabstMirror (approved):

PR #375, opened 2022-10-17
Normalize water floatation for vanilla heavy-lift helicopters to match CUP 

kil0byt3:
Make vanilla "Chinook" (aka Huron) and EH302 (IDAP/FIA heavy transport heli) float on water. 
This now matches CUP Chinook/CH-53e heavy lift helos. 

This change should eventually be extended to all helicopters from mods that could feasibly "float" (aka hover on surface of water for pseudo-helocast extract). I just don't currently have time to go through the editor and find all helos that should do this. I'll try to outsource it.

### DIFF
--- a/addons/miscFixes/CfgVehicles.hpp
+++ b/addons/miscFixes/CfgVehicles.hpp
@@ -50,6 +50,11 @@ class CfgVehicles {
         class TransportBackpacks {};
     };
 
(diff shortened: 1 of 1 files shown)


kil0byt3:
Can be merged and worked as-is (tested). However, can also be held back until all other modded helos are added.

PabstMirror (approved):

PR #412, opened 2023-06-27
OCAP implementation and compatibility with POTATO End Mission fnc

kil0byt3:
draft for now

### DIFF
--- a/addons/adminMenu/functions/fnc_endMission.sqf
+++ b/addons/adminMenu/functions/fnc_endMission.sqf
@@ -3,11 +3,9 @@
 params ["_winningSide"];
 TRACE_1("params",_this);
 
-private _missionType = getMissionConfigValue[QEGVAR(missionTesting, missionType), "other"];
-
-[_winningSide, "Mission ended", _missionType] call ocap_fnc_exportData;
-
 if (isServer) then {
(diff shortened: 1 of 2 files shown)


kil0byt3:
OCAP 2.0 automatically ends recording when it detects `BIS_fnc_endMission` or `BIS_fnc_endMissionServer`

I was able to implement without changing anything in POTATO, see link posted on Discord

There are more customizations we could make in this PR later, however, like adding tags to the missions i.e. TvT or Coop so they are sorted nicely on the web interface

I'll review that possibility after we test it for a bit

DanThomas64:
I think that we can close this now no?

kil0byt3:
> I think that we can close this now no?

I still want to do this: `adding tags to the missions i.e. TvT or Coop so they are sorted nicely on the web interface`

DanThomas64 (changes requested):
Suggest using 
`getMissionConfigValue QEGVAR(missionTesting missionType)` to get the type of mission. 

Technically The OCAP data would be saved in `endMission` and not `uihook_endMission` so I would create the debug message from the result of saving the OCAP recording based on the output of the ocap_fnc_exportData function.

TheCandianVendingMachine (approved):

PR #415, opened 2023-07-19
add player screams when they go unconscious

kil0byt3:
- I don't think we currently have a feature to do this but maybe
- it seems to double scream in singleplayer but I dont think it will in mp
- anecdotal evidence from various folks seems to think we don't have anything like this currently

### DIFF
--- a/addons/playerScreams/$PBOPREFIX$
+++ b/addons/playerScreams/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\potato\addons\playerScreams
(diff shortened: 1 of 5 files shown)


kil0byt3:
ready 4 merge

PabstMirror:
it oddly doesn't work for me
I don't think unconscious units can talk

e.g.
if I try
```
[] spawn { sleep 1;
cursorObject say3D [selectRandom potato_player_screams_soundList, 200, 1, false];
sleep 0.1;
[cursorObject, true, 60000] call ace_medical_fnc_setUnconscious;
};
```
the scream gets cut off after a fraction of a second

PabstMirror:
maybe switch to `playSound3D`?

kil0byt3:
I'll try that this week

kil0byt3:
Superseded with ACE Moaning

PR #416, opened 2023-07-19
remove "Init" in systemchat when opening arsenal

kil0byt3:
it's a tiny annoyance


addons/arsenalExport/functions/fnc_buttonClick.sqf
```diff
@@ -6,7 +6,6 @@ TRACE_2("canAttach",_display,_fncString);
 private _unit = missionNamespace getVariable ["ace_arsenal_center", player]; /// supports ace arsanal in 3den
 
 switch (toLower _fncString) do {
-case ("init"): {systemChat "Init";};
 case ("uniform"): {
```
PabstMirror, line 9:
```suggestion
case ("init"): {};
case ("uniform"): {
```
keep case or it falls back to the `default {ERROR_1("bad fnc [%1]",_fncString);};`

PR #418, opened 2023-08-01
RHSAFRF armor nerf 

kil0byt3:
Nerfs various RHS AFRF 6b armors to be more compatible with CUP and vanilla rounds (incl. POTATO ammo)

might be really inefficient way of doing this but I already had the configs sitting around


kil0byt3:
Probably no longer needed with latest ACE update

PR #419, opened 2023-08-04
Set LAMBS AI hiding off by default

kil0byt3:
currently set in Namespace. We should leave this PR in unmerged state until we've tested this setting in multiple Air-Mobile sessions to see how AI react


kil0byt3:
We have played many Air Mobile missions since I set this in Namespace, we are good to merge this in to default bwsettings

PabstMirror (approved):

PR #424, opened 2023-08-23
Resets BFT when SafeStart turns off. For missions using staged load-in (public events)

### LINKED ISSUE #411: Public Events: potato/bwmf BFT only works when all players load in to game at the same time
in the public event June 18, BFT markers didn't work for people loaded into the server at different times (we did a staged load-in)

fixed in public event by running 
`
if (hasInterface) then {
    ["potato_adminMenu_resetMarkers", [player]] call cba_fnc_localEvent;
};
`

maybe just set up another trigger to run that code when safestart gets turned off, for future public events

and/or add an admin menu button to fix bft

or something much more elegant, but those would work as a hacky fix that I can probably do

kil0byt3:
fixes #411 
untested, leaving draft for now

### DIFF
--- a/addons/safeStart/functions/fnc_toggleSafeStart.sqf
+++ b/addons/safeStart/functions/fnc_toggleSafeStart.sqf
@@ -26,4 +26,5 @@ if (_makeSafe) then {
 } else {
     missionNamespace setVariable [QGVAR(startTime_PV), -1, true];
     ["potato_safeStartOff"] call CBA_fnc_globalEvent;
+    if (hasInterface) then { ["potato_adminMenu_resetMarkers", [player]] call cba_fnc_localEvent; }; // Resets BFT for missions using staged load-in
 };


addons/safeStart/functions/fnc_toggleSafeStart.sqf
```diff
@@ -26,4 +26,5 @@ if (_makeSafe) then {
 } else {
     missionNamespace setVariable [QGVAR(startTime_PV), -1, true];
     ["potato_safeStartOff"] call CBA_fnc_globalEvent;
+    if (hasInterface) then { ["potato_adminMenu_resetMarkers", [player]] call cba_fnc_localEvent; }; // Resets BFT for missions using staged load-in
```
PabstMirror, line 29: this func only runs on server so it's fail the hasinterface
and potato_adminMenu_resetMarkers seems to want `player`, but I'm not sure why and that could probably be dropped

kil0byt3: is there a place you recommend to run this? I was going for the "dirty" method here and just making it reset markers when SafeStart is toggled, but really it would be more ideal if people's markers got a reset when they spawned (so they could see where PLT/COY is during SafeStart for briefing etc...). So maybe in a `PostInit` somewhere when players spawn? or even building something into the `markers` submod itself - let me know what your recommendation would be and I can build it from there

kil0byt3: disregard, see conversation in #scripting on discord, best catch-all is just to run it when safestart toggles

Juujuu:
Bump - If we have time to implement for public event, would be beneficial.

kil0byt3:
@DanThomas64 or @PabstMirror I might just ask you to fix this as I don't know if I have the time to "learn on the job" for this specifically

kil0byt3:
#434

PR #434, opened 2023-09-13
BFM Suggested fix

DanThomas64:
https://github.com/BourbonWarfare/POTATO/issues/411

### DIFF
--- a/addons/safeStart/XEH_postInit.sqf
+++ b/addons/safeStart/XEH_postInit.sqf
@@ -5,7 +5,6 @@ if (isServer) then {
     ["potato_safeStartOn", {
         if (!GVAR(syncFortify)) exitWith {};
         missionNamespace setVariable [QACEGVAR(fortify,fortifyAllowed), true, true];
-        call EFUNC(adminMenu, allPlayersResetMarkersButton);
     }] call CBA_fnc_addEventHandler;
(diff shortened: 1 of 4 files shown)


addons/safeStart/XEH_postInit.sqf
```diff
@@ -5,6 +5,7 @@ if (isServer) then {
     ["potato_safeStartOn", {
         if (!GVAR(syncFortify)) exitWith {};
         missionNamespace setVariable [QACEGVAR(fortify,fortifyAllowed), true, true];
+        call EFUNC(adminMenu, allPlayersResetMarkersButton);
```
PabstMirror, line 8:
```suggestion
        call EFUNC(adminMenu,allPlayersResetMarkersButton);
```

DanThomas64: I forget that is a thing. Good catch.

kil0byt3:
This should probably have been merged in before the public event, but we can just run the code manually at end of safestart for this one

kil0byt3:
This code didn't quite work during the CE; I am not sure if that was a problem with the code or just POTATO being overloaded by 100 players at the time

PabstMirror:
replaced with https://github.com/BourbonWarfare/POTATO/pull/631

PR #437, opened 2023-10-01
more bug fixes

PabstMirror:


addons/safeStart/functions/fnc_missionTimeWarning.sqf
```diff
@@ -17,23 +17,21 @@ TRACE_1("Params",_this);
...
 
-if (_missionLength == 0) then {
-    private _missionType = getMissionConfigValue QEGVAR(missionTesting,missionType);
-    switch (_missionType) do {
-        case 1: {_missionLength = 65};
-        case 2: {_missionLength = 35};
-        default {if(true) exitWith {TRACE_1("Something is seriously wrong, exiting timer",_missionType)}};
-    };
-};
+// if (_missionLength == 0) then {
```
PabstMirror, line 23: @DanThomas64 - this is a follow-up to https://github.com/BourbonWarfare/POTATO/pull/436/files
so as far as I can tell this hasn't worked in a while because of the nil, 
so I think it makes sense to disable this auto time for now
and just only run when explicitly set?

PR #443, opened 2023-10-17
3den - Rightclick menu to change rifleman to other types

PabstMirror:
i.e. easy change rifleman to additional aar
not sure if useful?

unit


kil0byt3:
i like this and think it's useful (don't see why not)

kil0byt3 (approved):

PR #444, opened 2023-10-19
Fix POTATO garrison'd units moving

TheCandianVendingMachine:
LAMBS `enableAI`'s `PATH` when requesting reinforcements so we simply give Zeus and option to just Not Do That. This obviously fixes all issues that i had while zeusing in my latest coop

New UI


kil0byt3 (approved):
LGTM thanks for testing

I thought I told you you're not allowed to do tech stuff :D

TheCandianVendingMachine:
bug fixes aren’t features :P

PR #447, opened 2023-10-22
feat: automatic Safestart off

### LINKED ISSUE #445: Safestart Auto-End Module
Blame Kilo, but it would be really nice to have a module that mission makers could place down that allows for them to configure how long safestart would last before it automatically ends. It's a nice QoL feature that some other frameworks have that I've missed but understand that it's not really that important :)

kil0byt3:
fixes #445 

backport from https://github.com/tanaaKa/POTATO/blob/788ad0cf7e6f6d1bba964aeaf4c0f68625f7cb48/addons/safeStart/XEH_postInit.sqf

adds an Eden toggle to "force" safestart, making it automatically turn off when time expires

also provides reminder at halfway through safestart

maybe we make this automatically turned on for TvTs somehow eventually

i haven't tested this yet, so draft

### DIFF
--- a/addons/missionTesting/CfgEden.hpp
+++ b/addons/missionTesting/CfgEden.hpp
@@ -118,6 +118,13 @@ class Cfg3DEN {
                             control = QUOTE(EditShort);
                             defaultValue = 0;
                             typeName = "NUMBER";
+                        };
+						class GVAR(ForceSS) {
+                            displayName = "Enforce safestart time:";
+							property = QGVAR(ForceSS);
+                            control = QUOTE(Checkbox);
+                            defaultValue = QUOTE(false);
+                            expression = "_this setVariable ['%s',_value];";
 						};
 						class GVAR(missionTimeLength) {
                             displayName = "Mission Length (mins):";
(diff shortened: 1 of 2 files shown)


PabstMirror:
some missions start events after SS going off
and our own end mission timer is based on when it goes off

so if there is some problem during safe start it would be good to delay it and that's something only humans can handle

how about just a reminder ping for admins
I would say at like 9:30 and 14:30 for how much time is left (so we don't turn off early if it's a 15miner)

AChesheireCat:
Very good point Pabst - I support the idea of a reminder ping for sure, regardless of if this change makes it in or not.

I still think this might be a nice thing to pick up as I _think_ the 3den "toggle" would help alleviate that issue by allowing MM's to choose whether or not they want the auto-off or not. If we do end up picking this change up, though, it might be a good idea to make sure that the toggle is set to "off" by default to prevent issues with safestart triggered events.

Just my 2cents.

kil0byt3:
agreed the reminder is what I want the most (mostly want the reminder for admins, but maybe no harm in the reminder going to everyone for Leadership Briefing's sake??) 

definitely should be off by default if we kept the Eden toggle

i can rework this a bit

kil0byt3:
good point re: tvts having sl highest rank

would add a bunch of complexity

they can just read their admin tab and we'll make this go to staff only :D

kil0byt3:
maybe a better way of doing this in the future, this solution is too complex for what's needed

PR #454, opened 2023-11-23
Adjust Smoke Plume near or on the fringes of water

DanThomas64:
Noted during testing of a mission on Isla Pera that even on land smoke was not filling as expected. 
This appeared to be a result of being near but not in water. 
These adjusted values change how the smoke behaves on the fringes of water. Any smoke created will plume above the water. And any smoke that is created underwater will not plume and fill, unless it is very very close to the surface. 

Example images from Isla Pera Before:

After:

After also tested in Tanoa:

### DIFF
--- a/addons/filledSmoke/config.cpp
+++ b/addons/filledSmoke/config.cpp
@@ -24,7 +24,7 @@ class CfgCloudlets {
         moveVelocity[] = {0,0.1,0}; //default {0,0.3,0}
         weight = 6.4; //default 1.26
         volume = 5; //default 1
-        destoryOnWaterSurface = 1;
+        destroyOnWaterSurface = 1;
         destroyOnWaterSurfaceOffset = 0.05; //default -0.3
         destroyAfterCrossing = "true";
         onSurface = "true";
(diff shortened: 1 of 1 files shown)


addons/filledSmoke/config.cpp
```diff
@@ -18,32 +18,17 @@ class CfgCloudlets {
...
-        // {0.7,0.7,0.7,0.1},
-        // {0.7,0.7,0.7,0.01}
-        // };
-        // randomDirectionPeriod= 0.3; //default 0.3
-        // randomDirectionIntensity = 0.15; //default 0.15
-        // destroyOnWaterSurfaceOffset = -0.3; //default -0.3
-        // MoveVelocityVar[] = {0.5,0.2,0.5}; //default {0.5,0.2,0.5}
-        // sizeVar = 0.6; //default 0.6
-        // colorVar[] = {0,0,0,0.05}; //default {0,0,0,0.05}
+        destoryOnWaterSurface = 1;
```
TheCandianVendingMachine, line 27: `destory` -> `destroy`
typo

TheCandianVendingMachine (changes requested):

TheCandianVendingMachine (approved):

PR #457, opened 2023-11-23
fix the fix of the mp7

kil0byt3:


addons/miscFixes/patchCUP/config.cpp
```diff
@@ -144,7 +144,7 @@ class CfgWeapons {
     };
 
     class Rifle_Base_F;
-    class Rifle_Short_Base_F; Rifle_Base_F
+    class Rifle_Short_Base_F: Rifle_Base_F;
```
TheCandianVendingMachine, line 147: ```class Rifle_Short_Base_F: Rifle_Base_F {};```

TheCandianVendingMachine (changes requested):

TheCandianVendingMachine (approved):

PR #464, opened 2023-12-03
Faux C7

AChesheireCat:
Adds a full-auto M16A1E1 and GL variant that is renamed to be "C7" and an M16A4 and GL variant that is renamed to "C7A2"


addons/fauxC7/CfgWeapons.hpp
```diff
@@ -0,0 +1,76 @@
...
+    };
+    class CUP_arifle_M16A1E1: CUP_arifle_XM16E1 {
+    };
+    class CUP_arifle_M16A1GL: CUP_arifle_XM16E1 {
+    }; 
+    class GVAR(c7): CUP_arifle_M16A1E1 {
+        author = "Potato";
+        displayName = "C7";
+        modes[] = {"Single","FullAuto","Burst_medium","single_medium_optics1","single_medium_optics2"};
+        class FullAuto: Mode_FullAuto {
```
TheCandianVendingMachine, line 25: why are we redefining this?

TheCandianVendingMachine (changes requested):

addons/fauxC7/CfgWeapons.hpp
AChesheireCat: IIRC, it was because the M16A1E1 was burst fire and so in order to change that to be more accurate, it had to be redefined... I think. Don't quote me on that lol.

TheCandianVendingMachine: Did you add this, or was it copy/pasted? As it is currently configured, the AI will go full-auto at a range of 300 meters

AChesheireCat: Copy/pasted from the CUP M16A1 config found in the same file, iirc.

TheCandianVendingMachine: Okay, did you test this in-game? I am not 100% sure this works, but I will need to investigate (if it doesnt work in game)

AChesheireCat: I have not tested this with AI - I didn't even think about it being used by an AI haha

It should work for players at least, all it did was inherit the existing model, renamed it, and converted it to full auto. It'd probably be a good idea to double check it tho

TheCandianVendingMachine: I am more talking about the inheritance of `Mode_fullAuto`; i don't know if this will be defined since the base gun is a 3 round burster

PabstMirror: I got this working, just needed to add sound configs

PabstMirror (approved):

PR #471, opened 2024-02-23
CUP_NVG_WP

AChesheireCat:
Adds White Phosphor (WP) versions of CUP NVGs that have IRL WP versions.


PabstMirror:
if it's ok, ace might just take this for their upcoming cup_compats?

AChesheireCat:
> if it's ok, ace might just take this for their upcoming cup_compats?

Yeah, feel free!

kil0byt3 (approved):
we will remove when cup_compats is live

PR #481, opened 2024-03-14
JSRS Compat + Class Structure fix

AChesheireCat:
- Adds JSRS sound compat, fixing the overly loud sound but replaces the unique sound with generic JSRS sound.

- PTV's last update changed the class structure for the weapons. This fixes the errors caused by that - specifically, the sights for the GL rifles being unusable as a result of this patch.


AChesheireCat:
The last patch, if added now, will probably bork the GL sights due to the class structure changeup.

PabstMirror (approved):

Issue #488, opened 2024-04-23
Project: True Viking "K-M/90 R GL" grenade launcher sight is broken.


AChesheireCat:
The K-M/90 R GL and G-M/90M GL's sight were broken by one of the PTV patches that I submitted; the specific problem is that when switching the firemode to the grenade launcher, it does not put the player's view to the GL sight and instead leaves it in the default "iron sight" view. As a result, it also prevents the player from ranging the GL and is restricted to whatever ranging the iron sights and/or the scope in use has.

Recommend reverting the GL compat patches that I made until a fix can be made that resolves the sight issue.

Bleh.

AChesheireCat:
For the record, this was reported by Stockton who found it out while testing - the issue appears limited to the K-M/90 R GL and the G-M/90M GL as far as I'm aware.

AChesheireCat:
Also to clarify, the sights work properly when no other mod is loaded, so it's definitely a me problem D:

kil0byt3:
Fixed in #489

PR #489, opened 2024-04-23
PTV GL Sight Fix + Structure Cleanup

AChesheireCat:
https://github.com/BourbonWarfare/POTATO/issues/488#issue-2257859724

This patch should fix whatever caused the GL sights to break on the K-M/90 R GL and the G-M/90M GL rifles.

This also brings the patch structure in-line with the current way that PTV restructured their classes (I think).

This should be verified, but I'm like 85% sure this works and fixes the issue at hand.

### DIFF
--- a/addons/miscFixes/patchPTV/config.cpp
+++ b/addons/miscFixes/patchPTV/config.cpp
@@ -41,18 +41,13 @@ class Mode_SemiAuto;
 class Mode_Burst;
 class Mode_FullAuto;
 class CfgWeapons {
+    class Default;    // External class reference
     class Pistol;
     class Pistol_Base_F: Pistol {};
     class Rifle;
     class Rifle_Base_F: Rifle {};
     class UGL_F;
-    class ptv_hk69_Base_F: Pistol_Base_F {};
-    // Adds compat for CUP GL ammunition in PTV GL weapons
-    class ptv_hk69: ptv_hk69_Base_F {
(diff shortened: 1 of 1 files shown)


kil0byt3 (approved):

addons/miscFixes/patchPTV/config.cpp
```diff
@@ -86,44 +82,34 @@ class CfgWeapons {
...
-            magazineWell[] = {"UGL_40x36","CBA_40mm_M203","CBA_40mm_EGLM"};
+                magazines[] = {"ptv_1Rnd_HE_Grenade_shell", __BIGL_MAGS, "CUP_1Rnd_HE_M203", "CUP_1Rnd_HEDP_M203", "CUP_1Rnd_StarCluster_White_M203", "CUP_1Rnd_StarCluster_Red_M203", "CUP_1Rnd_StarCluster_Green_M203", "CUP_1Rnd_StarFlare_White_M203", "CUP_1Rnd_StarFlare_Red_M203", "CUP_1Rnd_StarFlare_Green_M203", "CUP_FlareWhite_M203", "CUP_FlareGreen_M203", "CUP_FlareRed_M203", "CUP_FlareYellow_M203", "CUP_1Rnd_Smoke_M203", "CUP_1Rnd_SmokeRed_M203", "CUP_1Rnd_SmokeGreen_M203", "CUP_1Rnd_SmokeYellow_M203"};
```
PabstMirror, line 85: what is `__BIGL_MAGS`?
macro from pvt?

I think just setting the magwells should be enough

AChesheireCat: Yeah, __BIGL_MAGS was a macro from PTV iirc. 

And I think you're right about just the magazineWell being enough - that should just enable compat right?

PR #490, opened 2024-04-24
POTATO Spawn Module Unit Diversification

AChesheireCat:
Diversifies the rifleman units in modules that spawn more than one rifleman. Specifically, this affects the Patrol (2), Squad (8), and Dismount (6) modules.

This enables mission makers to use loadouts that have defined rifleman_02 and rifleman_03 kits, but does not affect the use of legacy loadouts that do not have rifleman_02 and rifleman_03 defined as those classes inherit from rifleman by default.

An example of this coming in handy is if a mission maker wishes to represent a guerilla faction that uses a wealth of weapons that do not share a common magwell. This will allow for the MM to use an M16A2 for the base rifleman kit, an FAL for rifleman_02, and an AKM for rifleman_03 and have them be represented in the spawn groups. This is currently not possible due to the groups spawning the base rifleman class three times instead.


TheCandianVendingMachine (approved):
lgtm

PR #496, opened 2024-05-21
RHS Medic Bag (Large)

AChesheireCat:
Adds a large version of the RHS Medic Bag that is sized to fit our medic's standard amount of supplies.

### DIFF
--- a/addons/miscFixes/patchRHSAFRF/config.cpp
+++ b/addons/miscFixes/patchRHSAFRF/config.cpp
@@ -17,7 +17,6 @@ class CfgPatches {
 
 class CfgWeapons {
     class ItemCore;
-    class B_AssaultPack_Base;
     class Vest_Camo_Base: ItemCore {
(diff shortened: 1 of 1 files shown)


addons/miscFixes/patchRHSAFRF/config.cpp
```diff
@@ -50,4 +51,10 @@ class CfgWeapons {
             };
         };
     };
+    class rhs_medic_bag: B_AssaultPack_Base {};
+    class rhs_medic_bag_XL: rhs_medic_bag {
```
PabstMirror, line 55: backpacks will be a cfgVeh, unless this is a vest? will check

AChesheireCat: You are, in fact, correct! Pushing a fix now.

PabstMirror (approved):

PR #498, opened 2024-05-22
CUP Weapons Reanim

AChesheireCat:
Updates a few CUP weapons to have newer/better animations. Specifically, the ACR has been changed to use the PTV HK433 animation while the old M4A1s have been updated to use the same animation as the "new" CUP M4A1s.

Weapons updated:

- ACR (5.56)
- ACR-C (5.56)
- ACR (6.8 SPC)
- ACR-C (6.8 SPC)
- M4A1 (11.3 MOE + RAS)
- M4A1 (14.5 MOE + RAS)


PabstMirror:
looks good, I just want to move it to a separate patch that will handle requiring both pvt and cup

PabstMirror (approved):

PR #500, opened 2024-05-29
fix weapon lowering while walking weird animation

kil0byt3:
(current)
when you lower your weapon with double ctrl, you'll be locked into moving in the direction you were traveling for a few seconds, even if you stop giving movement inputs.

(when fix is applied)
when you lower your weapon with double ctrl, and if you stop giving movement inputs, you'll stop moving within half of a second

(limitations):
if you continue giving movement inputs, you'll keep moving as if the fix was not applied. I don't know if there's anything that can be done about this

(intent)
make movement feel a bit more responsive. It takes a bit of getting used to _removing_ your hand from the keyboard to stop, but it works a lot better than the current system in which you can't stop moving once the animation is playing

if you want to test it yourself you can uncomment the two `systemChat` lines and it'll print when it applies


kil0byt3:
tested, this works,

(current)
when you lower your weapon with double ctrl, you'll be locked into moving in the direction you were traveling for a few seconds, even if you stop giving movement inputs.

(when fix is applied)
when you lower your weapon with double ctrl, and if you stop giving movement inputs, you'll stop moving within half of a second

(limitations):
if you continue giving movement inputs, you'll keep moving as if the fix was not applied. I don't know if there's anything that can be done about this

(intent)
make movement feel a bit more responsive. It takes a bit of getting used to _removing_ your hand from the keyboard to stop, but it works a lot better than the current system in which you can't stop moving once the animation is playing

if you want to test it yourself you can uncomment the two `systemChat` lines and it'll print when it applies

added the above summary to the PR and marked as ready for review

addons/miscFixes/XEH_preInit.sqf
```diff
@@ -22,3 +22,53 @@ if (["WBK_ZombieCreatures"] call ACEFUNC(common,isModLoaded)) then {
...
+        _nAnim set [11,"p"];
+        _nAnim set [21,"n"];
+        _nAnim set [22,"o"];
+        _nAnim set [23,"n"];
+        _nAnim = _nAnim joinString "";
+        player playMoveNow _nAnim;
+        if(isMultiplayer)then{
+            [player,_nAnim] remoteExec {
+                params ['_u','_nAnim'];
+                _u playMove _nAnim;
```
PabstMirror, line 68: playMoveNow should have global effects, so I don't think we need this

PabstMirror (approved):

PR #501, opened 2024-06-01
ACE Wake Up for Non-stable Patients

AChesheireCat:
Adds a flat, non-adjustable wake-up chance for unconscious and non-stable player patients. This is intended to decrease the brutality of going uncon and having to wait for your teammates to check on you.

Wants for this feature:

- Add a %chance to go back unconscious if the patient is not stable and/or has low blood volume. This will highly incentivize players to receive medical attention before running off despite stabilizing themselves.

I've written some pseudo-code so as to show what the above concept might look like.

### DIFF
--- a/addons/aceWakeup/config.cpp
+++ b/addons/aceWakeup/config.cpp
@@ -1,0 +1,24 @@
+#include "script_component.hpp"
+
+class CfgFunctions {
+    class overwrite_ace_medical_status {
+        tag = "ace_medical_status";
+        class ace_medical_status {
+            class hasStableVitals {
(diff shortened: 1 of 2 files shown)


AChesheireCat:
To note: the values currently in place were tested against the base BW ACE Medical settings, but are not set in stone. Per testing done locally, the current values result in the average wake up time for units below class 2 hemorrhage being about 1minute, whereas units below class 4 hemorrhage having an average wake up time of 4 minutes.

PabstMirror:
I'm worried people will be able to wake up too quickly in tvts
I'd like to think about other ways of doing this, maybe limiting to coops only

PR #511, opened 2024-07-30
POTATO RPG29 Patch

AChesheireCat:
This patch removes the RHSUSAF dependency for the RPG 29 Vampir mod by replacing the ammunition configs with RHSAFRF and CUP equivalents.

*It should be noted that this code is really messy and I'm pretty sure some of it is redundant, but it works and I'm very scared to remove anything lest it breaks...

### DIFF
--- a/addons/miscFixes/patchRPG29/config.cpp
+++ b/addons/miscFixes/patchRPG29/config.cpp
@@ -1,0 +1,357 @@
+#include "\z\potato\addons\miscFixes\script_component.hpp"
+#undef COMPONENT
+#define COMPONENT miscFixes_patchRPG29
+
+class CfgPatches {
+    class ADDON {
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = { "RPG29", "potato_core", "rhs_main_loadorder" };
+        skipWhenMissingDependencies = 1;
+        author = "Bourbon Warfare";
+        authorURL = "https://github.com/BourbonWarfare/POTATO";
+        VERSION_CONFIG;
+    };
+};
+
+class Mode_SemiAuto;
+class Mode_Burst;
+class Mode_FullAuto;
+class CowsSlot;
+class MuzzleSlot;
+class SlotInfo;
+class WeaponSlotsInfo;
+class asdg_SlotInfo;
+class asdg_OpticRail;
+class asdg_OpticRail1913;
+class asdg_MuzzleSlot;
+class asdg_MuzzleSlot_338;
+class PointerSlot;
+class asdg_FrontSideRail;
+class asdg_UnderSlot;
+class BaseSoundModeType;
+class CfgMagazineWells {
+	class RPG29 {
+		BI_Magazines[] = {"RPG29_AT","RPG29_HE_F","potato_RPG29_FRAG"};
+	};
+};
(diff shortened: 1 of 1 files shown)


TheCandianVendingMachine:
which classes were causing the dependency? looking at this, this is just recreating the entire weapon config from scratch. you say its just ammo, but it touches a lot

AChesheireCat:
The dependencies were just the ammo (it inherited from the RHSUSAF MAAWS rounds), but I also went a little out of scope to add a scratch-made version of the RPG-29's OG-29 round that inherits from the CUP RPG-7V OG-7V round. 

I'd like to reiterate that this code is *really* messy - every time I touched a different part of it when making the patch seemed to break something else and I've already put too many hours into making this patch so once I got it into a solid, reliable working state, I pushed it to at least get eyes on it. I'll probably continue to refine it, but at a much later date.

kil0byt3 (approved):

PabstMirror:
maybe just reduce this to just what we need to change


this replaces the rhsus sounds
and then the `ammo` with the vanilla maaws
just need to find something for the frag round
```

class CfgMagazineWells {
    class RPG29 {
        BI_Magazines[] = {"RPG29_AT","RPG29_HE_F","potato_RPG29_FRAG"};
    };
};

class CfgMagazines {
    class CA_LauncherMagazine;
    class RPG29_AT: CA_LauncherMagazine {
        ammo = "R_MRAAWS_HEAT_F";
    };
    class RPG29_HE_F: CA_LauncherMagazine {
        ammo = "R_MRAAWS_HE_F";
    };
    class potato_RPG29_FRAG: CA_LauncherMagazine { // OG-29 HE-FRAG round
        // ammo = "?";
    };
};

class Mode_SemiAuto;
class CfgWeapons {
    class Launcher_Base_F;
    class CUP_launch_MAAWS: Launcher_Base_F {};
    class RPG29_Lorad: CUP_launch_MAAWS {
        class Single: Mode_SemiAuto {         
            class StandardSound {
                begin1[] = {"A3\Sounds_F\arsenal\weapons\Launchers\RPG32\rpg32",1.9952624,1,1500}; // replace rhsus sound
                soundBegin[] = {"begin1",1};
            };
        };
        reloadMagazineSound[] = {"A3\Sounds_F\arsenal\weapons\Launchers\RPG32\reload_RPG32",0.25118864,1,10}; // replace rhsus sound
    };
};
```

AChesheireCat:
> maybe just reduce this to just what we need to change
> 
> this replaces the rhsus sounds and then the `ammo` with the vanilla maaws just need to find something for the frag round

Yeah, that's what I had tried initially because that was the easiest solution to implement, but it resulted in weird bugs like the round not being visualized in the launcher when loaded, the rocket shooting in a direct straight line with no drop, or the rocket doing the CoD4 RPG thing where it spirals randomly. The only solution that seemed to work was to transpose all of the original ammunition's values and replace some of it with the CUP equivalent values. 

Although, there is a lot of left over code from when I was experimenting. I really do have to sit down eventually to pare down and clear it all up. That probably won't be for another week or so, though.

TheCandianVendingMachine:
tabs -> spaces?

lambdatiger:
Since Chesh did the work of making a mod for this (ty 😘), do we need this PR still? Or are we planning to close it?

PR #513, opened 2024-08-02
Fixes Crash-To-Lobby bug caused by PLA

AChesheireCat:
PLA updated their config syntax/class names recently which broke this patch and caused a Crash-to-Lobby when the ZTL-11 and ZTD-05 were present in a mission.

This *should* resolve that issue.


PabstMirror (approved):
placed ztd/ztl without any popups

PR #463, opened 2023-12-03
PLA Armor Normalization

AChesheireCat:
This patch is intended to normalize the PLA Armored Vehicles mod's armor to vanilla standards so as to fix the ongoing issue where infantry AT and equivalent vehicles cannot do meaningful damage to the PLA vehicles.

Specifically, this patch affects the following vehicles:

- ZBL09
- ZTL11
- ZBD04A
- ZBD05
- ZTD05

PR #515, opened 2024-08-05
PLA Tank Patch

watermelon-0302:
A continuation of Chesh's PLA Armoured Vehicles fix in this PR: https://github.com/BourbonWarfare/POTATO/pull/463

This PR normalizes tank values for the ZTQ15, ZTQ15B, ZTZ96A, ZTZ96B, ZTZ99 and ZTZ99A to the CUP T-72, T-90 and T-90M. Specifically:

- The ZTQ15 and ZTQ15B have their values normalized to a CUP T-72;
- The ZTZ96A and ZTZ96B have their values normalized to a CUP T-90;
- The ZTZ99 and ZTZ99A have their values normalized to a CUP T-90M;


addons/miscFixes/patchPLA/config.cpp
AChesheireCat, line 1: I'm concerned with a negative value being used here - typically that's only used for parts that have their passThrough value set to 0 and are non-essential parts like tracks/tires.

AChesheireCat, line 1: These values are not being changed, right? If so, they can be excised without breaking the patch.

AChesheireCat, line 1: Same for these

AChesheireCat, line 1: Extraneous inclusion?

AChesheireCat, line 1: Breaking inheritance runs the risk of breaking a lot of stuff. I'm not comfortable with doing this without this being tested extensively.

AChesheireCat, line 1: Same for this.

AChesheireCat (dismissed):

addons/miscFixes/patchPLA/config.cpp
watermelon-0302: if i'm understand scope correctly, the classes should inherit from the same file and not from the original base addon. but if it is a genuine concern and to avoid bugs, i'll push a change to make everything explicit.

watermelon-0302: gotcha, i'll go ahead and put any negative values i find to 0 to prevent wonkyness, at least for components which have their passThrough value set to anything other than 0.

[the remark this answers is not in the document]
watermelon-0302: what vehicles do you suggest? will the Black Eagle and the Armata work well?

watermelon-0302: yes. did not clean the file up properly.

AChesheireCat:
Request Review pl0x

kil0byt3 (approved):

addons/miscFixes/patchPLA/config.cpp
```diff
@@ -5,19 +5,20 @@
...
-        authors[] = {"Chesheire"};
+        authors[] = {"Chesheire","watermelon"};
         authorUrl = "https://github.com/BourbonWarfare/POTATO";
         VERSION_CONFIG;
     };
 };
 
 class CfgVehicles {
-    class Car;
+    class LandVehicle;
```
PabstMirror, line 20: duped on line 50

AChesheireCat: Good catch, fix'd!

kil0byt3:
testing and working as expected

PabstMirror (approved):

PR #521, opened 2024-08-18
Add GM Missile Guidance Compat and other small fixes

lambdatiger:
In this PR, I propose changes to

- Adds compats for GM missiles/rockets referencing TCVM's other work within this addon and in ACE
  - There is a wild stab at the Milan parameters in this PR.
- Add missile guidance compatibility for the CUP dragon 
- Override JSRS's soundset so the distance pops of the Dragon don't sound like large explosions.
- Accidentally tab->spaces the CWR compat
 
Posting this for feedback.

### TODO
- [x] Find Milan parameter ~~reference and update w/ that.~~ that seem fun.

### DIFF
--- a/addons/missileGuidanceCompat/CfgAmmo.hpp
+++ b/addons/missileGuidanceCompat/CfgAmmo.hpp
@@ -4,6 +4,11 @@ class CfgAmmo {
     class M_Titan_AT;
     class Bo_GBU12_LGB;
 
+    class ShellBase;
+    class ace_dragon_serviceCharge: ShellBase {
+        soundSetExplosion[] = {};
+    };
+
     class CUP_M_9K32_Strela_2_AA: M_Titan_AA {
         maneuvrability = 0;
         #include "CfgMissileStrela.hpp"
--- a/addons/missileGuidanceCompat/CfgMissileHOT.hpp
+++ b/addons/missileGuidanceCompat/CfgMissileHOT.hpp
@@ -1,0 +1,38 @@
+class ace_missileguidance { // reference https://github.com/acemod/ACE3/pull/10029
+    enabled = 1;
+
+    pitchRate = 45;          // Minium flap deflection for guidance
+    yawRate = 45;            // Maximum flap deflection for guidance
+
+    canVanillaLock = 0;          // Can this default vanilla lock? Only applicable to non-cadet mode
+
+    showTrail = 1;
+
+    // Guidance type for munitions
+    defaultSeekerType = "SACLOS";
+    seekerTypes[] = { "SACLOS" };
+
+    defaultSeekerLockMode = "LOAL";
+    seekerLockModes[] = { "LOAL", "LOBL" };
+
+    defaultNavigationType = "Line";
+    navigationTypes[] = { "Line" };
+
+    lineGainP = 7;
+    lineGainD = 6;
+
+    initialPitch = 2;
+
+    seekLastTargetPos = 0;      // seek last target position [if seeker loses LOS of target, continue to last known pos]
+    seekerAngle = 30;           // Angle from the shooter's view that can track the missile
+    seekerAccuracy = 1;         // seeker accuracy multiplier
+
+    seekerMinRange = 75;
+    seekerMaxRange = 4000;      // Range from the missile which the seeker can visually search
+
+    offsetFromCrosshair[] = { 0, 0, 0 }; // where the missile wants to stay in relation to the center of the crosshair.
+
+    // Attack profile type selection
+    defaultAttackProfile = "WIRE";
+    attackProfiles[] = {"WIRE"};
+};
--- a/addons/missileGuidanceCompat/patchGM/config.cpp
+++ b/addons/missileGuidanceCompat/patchGM/config.cpp
@@ -61,7 +61,8 @@ class CfgAmmo {
             pitchRate = 25;
             yawRate = 25;
             initialPitch = 0;
-            lineGainD = -3;
+            lineGainP = 7;
+            lineGainD = 4;
         };
     };
     // HOT missile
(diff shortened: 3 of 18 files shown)


kil0byt3:
This is a based PR

I'll review when I am at my PC tomorrow but ty for all of the work.

lambdatiger:
I'm happy with how it is, try it out and provide any feedback you might have.

addons/missileGuidanceCompat/patchGM/config.cpp
```diff
@@ -0,0 +1,157 @@
...
+        #include "../CfgMissileBastion.hpp"
+    };
+    class gm_missile_bastion_heat_9M117: gm_missile_bastion_base {
+        class ace_missileguidance: ace_missileguidance {
+            enabled = 1;
+            pitchRate = 25;
+            yawRate = 25;
+            lineGainP = 6;
+            lineGainD = 5;
+            showTrail = 0;
```
TheCandianVendingMachine, line 39: generally I enable `showTrail` for player guided missiles to see them at a distance, not blocking though

addons/missileGuidanceCompat/CfgMissileHOT.hpp
```diff
@@ -0,0 +1,38 @@
...
+    initialPitch = 2;
+
+    seekLastTargetPos = 0;      // seek last target position [if seeker loses LOS of target, continue to last known pos]
+    seekerAngle = 30;           // Angle from the shooter's view that can track the missile
+    seekerAccuracy = 1;         // seeker accuracy multiplier
+
+    seekerMinRange = 75;
+    seekerMaxRange = 4000;      // Range from the missile which the seeker can visually search
+
+    offsetFromCrosshair[] = { 0, 0, 0 }; // where the missile wants to stay in relation to the center of the crosshair.
```
TheCandianVendingMachine, line 33: I don't have a source for this, but the HOT missile flies 0.3 meters above the crosshair. So I think it should be `{ 0, 0, 0.3 }` for perfect translation. also not blocking

addons/missileGuidanceCompat/CfgAmmo.hpp
```diff
@@ -4,6 +4,11 @@ class CfgAmmo {
     class M_Titan_AT;
     class Bo_GBU12_LGB;
 
+    class ShellBase;
+    class ace_dragon_serviceCharge: ShellBase {
+        soundSetExplosion[] = {};
```
TheCandianVendingMachine, line 9: I don't know much about ARMA audio, but make sure that a player can hear the dragon flying towards them. I'll accept after it's tested

TheCandianVendingMachine (changes requested):

addons/missileGuidanceCompat/patchGM/config.cpp
lambdatiger: GM adds some stuff to their models to make them fairly visible, I'll post pictures after work.

addons/missileGuidanceCompat/CfgMissileHOT.hpp
lambdatiger: Is that a optic offset from the beam projector, or is that how the missile sits in the beam? It doesn't seem there's an offset in [this video](https://www.youtube.com/watch?v=tZQaqCNj4mM), but it also looks like the parameters for the HOT-3 I just copy and pasted may not work the way the missile does.

addons/missileGuidanceCompat/CfgAmmo.hpp
lambdatiger: JSRS adds it's sound sets as it's an easy way to override / customize sound. The soundset would normally be undefined, but with JSRS inserts an entry. With JSRS's entry, at a distance you hear JSRS medium-sized explosions instead of the actual service charge pops which is confusing to say the least. Again, video when I get off work

lambdatiger: First half is w/o the fix, second half is with the fix
https://github.com/user-attachments/assets/75555729-a1a2-4825-806f-2c6e703c4872

addons/missileGuidanceCompat/CfgMissileHOT.hpp
lambdatiger: > it doesn't seem there's an offset in [this video](https://www.youtube.com/watch?v=D2jIgR7twmc), but it also looks like the parameters for the HOT-3 I just copy and pasted from your PR may not work the way the missile does (or at least the HOT-1).

Looking at that video again, and some other specs on the HOT missile, I think this is a second helicopter watching it go in

addons/missileGuidanceCompat/patchGM/config.cpp
TheCandianVendingMachine: yeah im good with it in this case then, ill look at the images but i trust it's decent

addons/missileGuidanceCompat/CfgMissileHOT.hpp
TheCandianVendingMachine: HOT3 has the offest, not HOT1/HOT2. This is fine

addons/missileGuidanceCompat/CfgAmmo.hpp
TheCandianVendingMachine: Some people will miss the soundscape; but I am fine with it either way. I'm resolving it for now, I was just worried that the audio would be completely messy

TheCandianVendingMachine (approved):

addons/missileGuidanceCompat/patchGM/config.cpp
lambdatiger: on discord in scripting

TheCandianVendingMachine:
still approved

PabstMirror (approved):

addons/missileGuidanceCompat/CfgAmmo.hpp
lambdatiger: I can edit together a video that's clearer on the changes and submit a PR if it should.

PR #523, opened 2024-08-23
RHS Body Armor Normalization

AChesheireCat:
Brings the RHS vest armor values to be more in line with Vanilla/CUP as well as to clarify the armor level rating system which is labelled after the Russian GOST system, not NIJ like the vanilla/CUP armor systems.

This affects the following RHS AFRF vests and their subvariants:

- 6B2
- 6B3
- 6B5
- 6B13
- 6B23

### DIFF
--- a/addons/miscFixes/patchRHSAFRF/config.cpp
+++ b/addons/miscFixes/patchRHSAFRF/config.cpp
@@ -50,6 +50,305 @@ class CfgWeapons {
             };
         };
     };
+    class rhs_6b23: Vest_Camo_Base {
+        descriptionShort = "GOST Armor Level 4"; // was "Armor Level 4"
+        class ItemInfo: ItemInfo {
+            class HitpointsProtectionInfo {
+                class Neck {
+                    HitpointName = "HitNeck";
+                    armor = 8;
+                    PassThrough = 0.4;
+                };
+                class Chest {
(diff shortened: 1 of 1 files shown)


kil0byt3 (approved):
I thought the issue with RHS armors was `passThrough `values

Do those need any tweaking or is just modifying `armor`enough?

Regardless lg2m as-is

AChesheireCat:
> I thought the issue with RHS armors was `passThrough `values
> 
> Do those need any tweaking or is just modifying `armor`enough?
> [...]

Sorta-kinda. Certain armors had their pass-through values tweaked (ie. the 6B13 had a pass-through of 0.1 which is extremely tanky) but others were in line with vanilla values. CUP themselves use 0.3 for their equivalent armors.

The real big issue when I looked into it did seem to be the armor values - using the 6B13 as an example again, it had 28 listed as the armor value. Comparatively, the most modern Russian CUP vest, the 6B45, used *16* as their armor value. Even if the pass-through value had been upped to allow more damage through, the threshold to reach is extremely high. I chose the easy route and just normalized the values to avoid having to do tiny specific tweaks and shiz.

kil0byt3:
Cool. Good info. Sounds good!

PabstMirror (approved):

PR #524, opened 2024-08-23
Arsenal Export - Hide when ace_arsenal is hidden

### LINKED ISSUE #503: POTATO loadout setter in Arsenal; doesn't hide when pressing Backspace
title, also would be nice to have a way to move the UI window around, minimize it, or otherwise hide it.

Hiding it by default or having a POTATO CBA setting to choose whether it's shown/hidden would be nice too, so that members can do Offnight missions that use the Arsenal, without having to look at the UI elements. 

Mission Makers could have the CBA option on, and always see the loadout setter UI in the arsenal for mission making purposes

PabstMirror:
Close #503

### DIFF
--- a/addons/arsenalExport/functions/fnc_arsenalOpened.sqf
+++ b/addons/arsenalExport/functions/fnc_arsenalOpened.sqf
@@ -22,6 +22,15 @@ private _yPos = (safezoneY + 1.5 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 
 
 private _ctrlGroup = _display ctrlCreate ["RscControlsGroup", IDC_CTRLGROUP];
 
+private _fnc_updateInfo = {
+    params ["_display"];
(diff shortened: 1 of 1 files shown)


addons/arsenalExport/functions/fnc_arsenalOpened.sqf
```diff
@@ -22,6 +22,15 @@ private _yPos = (safezoneY + 1.5 * ((((safezoneW / safezoneH) min 1.2) / 1.2) /
 
 private _ctrlGroup = _display ctrlCreate ["RscControlsGroup", IDC_CTRLGROUP];
 
+private _fnc_updateInfo = {
+    params ["_display"];
+    private _aceCtrlMenu = _display displayCtrl 10;
+    private _ctrlGroup = _display displayCtrl IDC_CTRLGROUP;
+    if (isNull _aceMenuCtrl) exitWith {};
```
PabstMirror, line 29: var name ☹️

Issue #532, opened 2024-08-28
Allow CUP Mk16 EGLM and Mk16 CQC EGLM to use NIArms and PTV STANAG magazines


AChesheireCat:
It would be nice if the CUP Mk16 EGLM and it's CQC variant were able to use the NIArms and PTV 30 and 40 round magazines, so that way you don't need to have two different magazine defines for regulars and leaders, or being forced to use only the CUP magazines when using the Mk16.

PR #533, opened 2024-08-28
Add NIArms and PTV STANAG magazines to the `CBA_556x45_SCAR_EGLM` Magwell 

lambdatiger:
This PR adds PTV and NIA magazines to the `CBA_556x45_SCAR_EGLM` magwells, addressing https://github.com/BourbonWarfare/POTATO/issues/532.

Note there are a few NI Arms magazines that clip the magwell slightly if you look directly at it. Chesh requested that we keep the ones with slight visual clipping in.


PabstMirror (approved):

PabstMirror:
reason they have different magwells

> SCAR-L with EGLM/FN40GL/Mk 13 Mod 0 grenage launcher (This extends the length of the magwell and prevents mags that are too short or wide from being used)

but I think for us it's fine as it just makes it easier to use

lambdatiger:
We (Chesh and I) went through and removed all the ones that very obviously wouldn't fit, there are a few NI Arms mags where the model slightly clips into the magwell (you have to be looking at the magwell closely to notice) that Chesh requested to stay.

lambdatiger:
Here's the mags this PR adds

https://github.com/user-attachments/assets/b28e065e-8067-4edf-90c8-3f609a893633

PR #535, opened 2024-08-31
GM Placeable Bridges 

lambdatiger:
This PR would add editor and zeus placeable bridges from the GM cDLC. Same content as [this mod](https://steamcommunity.com/sharedfiles/filedetails/?id=3310038876), but without the editor previews.


lambdatiger:
Going to fix how I branched this and then remake this PR

PR #540, opened 2024-09-07
Add ability to create larger supply boxes with smaller resupply boxes inside them

lambdatiger:
This adds an idea kilo had/stole in the [resupply boxes suggestion on BW's discord](https://discord.com/channels/204621032428929025/1281258900728975382/1281672958061838498). It allows a mission maker to use the `CfgLoadouts` `SupplyBoxes` class to make larger boxes with multiple smaller resupply boxes inside them. See this example
```cpp
class CfgLoadouts {
  // other code before or after
  // Do Supply Box Loadouts
  // (1 will run normaly, 0 will leave them to vanilla defaults, -1 will clear and leave empty)
  setSupplyBoxLoadouts = 1;
  class SupplyBoxes {
    class ACE_medicalSupplyCrate {
      boxCustomName = "Wild Medical Box";
      TransportItems[] = {
        "ACE_elasticBandage:40",
        "ACE_packingBandage:20",
        "ACE_epinephrine:15",
        "ACE_morphine:15",
        "ACE_adenosine:15",
        "ACE_tourniquet:10",
        "ACE_splint:10",
        "ACE_salineIV:4",
        "ACE_salineIV_500:4",
        "ACE_salineIV_250:10"
      };
    };
    class I_E_CargoNet_01_ammo_F {
       // Add ability to mark the larger supply boxes
       // 0 for no actions
       // 1 for 1x smoke and 1x glow stick
       // 2 for 1x smoke and 1x glow stick, and a quick map marker based off of box name
      addMarkingActions = 2;
      // number of boxes that can be put in the crate, default is four,
      // but should be more than the number of sub boxes desired
      boxCustomName = "Multi Box Drifting";
      boxSpace = 4;
      class Box_NATO_Ammo_F { // doesn't affect configuration (or lack thereof) for this box elsewhere
        boxCount = 2; // default is 1 box
        boxCustomName = "Resupply Box";
        TransportWeapons[] = {"gm_m72a3_oli:2"};
        TransportMagazines[] = {
          "CUP_30Rnd_556x45_Stanag:30",
          "CUP_30Rnd_556x45_Stanag_Tracer_Red:6",
          "CUP_100Rnd_TE4_LRT4_Red_Tracer_762x51_Belt_M:8",
          "CUP_1Rnd_HE_M203:6",
          "CUP_1Rnd_HEDP_M203:4",
          "HandGrenade:8",
          "SmokeShell:5",
          "SmokeShellGreen:1",
[...]

PR #539, opened 2024-09-06
add ability to convert vehicle loadouts into crates

TheCandianVendingMachine:
requested by kilo, vehicle loadouts will automatically be added into ACE cargo as ammo boxes with appropiate number of equipment

https://youtu.be/1A0rzEVLtgg?si=5lEhtpDXDP2Y1sAG

PR #544, opened 2024-09-10
Add ability to define vehicle loadouts to be stored in crates

lambdatiger:
This lets you add supply boxes in a similar fashion to the changes added in #540, but for vehicles. Similar parameters as before, but in this case both the normal `TransportXXX[]` arrays and the classes can be present, and the classes are only used in `setVehicleLoadouts = 3`, ignoring the other storage arrays.

```cpp
class Car {
  TransportWeapons[] = {AT};
  TransportMagazines[] = {RIFLE_MAG,GLRIFLE_MAG,CARBINE_MAG,AR_MAG,AR_MAG,GLRIFLE_MAG_HE};
  TransportItems[] = {BASE_MEDICAL,BASE_MEDICAL,BASE_MEDICAL,BASE_MEDICAL};
  class Box_NATO_Ammo_F {
    boxCount = 2;
    boxCustomName = "Resupply Box";
    TransportWeapons[] = {AT};
    TransportMagazines[] = {
      RIFLE_MAG, RIFLE_MAG, RIFLE_MAG,
      GLRIFLE_MAG, CARBINE_MAG, AR_MAG, GLRIFLE_MAG_HE
    };
    TransportItems[] = {
      "ACE_elasticBandage:25",
      "ACE_packingBandage:15",
      "ACE_tourniquet:5",
      "ACE_splint:5"
    };
  };
};
class gm_ge_army_u1300l_cargo { // specific logi vehicle
  minVehicleBoxSpace = 20; // the new amount of cargo space
  class Box_NATO_Ammo_F {
    boxCount = 12; // the number of boxes of this type to put into the vehicle
    boxCustomName = "Resupply Box";
    TransportWeapons[] = {AT};
    TransportMagazines[] = {
      RIFLE_MAG, RIFLE_MAG, RIFLE_MAG,
      GLRIFLE_MAG, CARBINE_MAG, AR_MAG, GLRIFLE_MAG_HE
    };
    TransportItems[] = {
      "ACE_elasticBandage:25",
      "ACE_packingBandage:15",
      "ACE_tourniquet:5",
      "ACE_splint:5"
    };
  };
};
```
This requires #539 and #540 to be merged first, and then some merge conflicts to be dealt with.


addons/assignGear/functions/fnc_assignGearVehicle.sqf
```diff
@@ -76,50 +80,279 @@ if (!isClass _path) exitWith {
...
+    case 3: { // ammo in boxes in vehicle from config
+        private _boxes = "true" configClasses _path;
+        {
+            private _boxType = configName _x;
+            private _boxCount = (getNumber (_x >> "boxCount")) max 1;
+            for "_i" from 1 to _boxCount do {
+                private _box = createVehicle [_boxType, [2,5,105], [], 0, "CAN_COLLIDE"];
+                [_box, _x, ["%1", "%1 " + str _i] select (_boxCount > 1)] call FUNC(setBoxContentsFromConfig);
+                [_box, 1] call ACEFUNC(cargo,setSize);
+                if !([_box, _theVehicle, true] call ACEFUNC(cargo,loadItem)) exitWith {
```
TheCandianVendingMachine, line 343: I'm not a fan of mission makers needing to set each vehicle cargo size to accommodate this: loadout files should be plug and play across any mission without needing to enter the editor at all. Why not set the box size to zero, or adjust the vehicle cargo space to accommodate the box?

addons/assignGear/functions/fnc_setContainerContentsFromConfig.sqf
```diff
@@ -0,0 +1,61 @@
+#include "script_component.hpp"
+/*
+ * Author: Bailey
```
TheCandianVendingMachine, line 3: I'm not the author of this code, credit `PabstMirror`

addons/assignGear/functions/fnc_setBoxContentsFromConfig.sqf
```diff
@@ -0,0 +1,34 @@
+#include "script_component.hpp"
+/*
+ * Author: Bailey
```
TheCandianVendingMachine, line 3: I don't think I'm the author of this?

TheCandianVendingMachine (changes requested):

addons/assignGear/functions/fnc_setBoxContentsFromConfig.sqf
lambdatiger: [This is your code](https://github.com/BourbonWarfare/POTATO/blob/b1bf31beac2bc598f6fe0c6847a603612b908a38/addons/assignGear/functions/fnc_assignGearSupplyBox.sqf#L44), I just moved it to a function. If you want, I can throw my name on there too, but I didn't really write any of it, I just checked your work

addons/assignGear/functions/fnc_setContainerContentsFromConfig.sqf
lambdatiger: Gotcha, is that true for the other comment as well?

addons/assignGear/functions/fnc_setBoxContentsFromConfig.sqf
TheCandianVendingMachine: Oh yep, I thought it was for vehicles, not boxes

addons/assignGear/functions/fnc_setContainerContentsFromConfig.sqf
TheCandianVendingMachine: Just this

addons/assignGear/functions/fnc_assignGearVehicle.sqf
lambdatiger: Preface this with my personal preference and choice would be just to leave it as is, I think shrinking the size of each container down to 1 is good enough for a mission maker to put boxes in. Especially considering you're probably only putting 3-4 boxes in a truck per coop at most (obviously logi withstanding, but their trucks/vehicle can usually handle more).

> Im not a fan of mission makers needing to set each vehicle cargo size to accommodate this
I can set a hook so they can do similar to #540, where they set the cargo size by config, or would you also like changes there as well? 
> Why not set the box size to zero
I'm not a fan instant put in / take out for boxes and it allows you to take all resupply boxes from a immobile vehicle and pile on in instead of making choices.
> adjust the vehicle cargo space to accommodate the box?
Again I could add a what I mentioned above, but by and large I think vehicle ACE cargo spaces are accurate and reasonable to what should be expected of a vehicle.

addons/assignGear/functions/fnc_setBoxContentsFromConfig.sqf
lambdatiger: Wait if this is and the other isn't, what's the difference in the code I'm missing? I think I changed the variable name to `_theContainer` in the other.

addons/assignGear/functions/fnc_assignGearVehicle.sqf
TheCandianVendingMachine: I agree the insta-unloading is not favourable, but this does introduce barriers to mission makers. A goal of POTATO is to let mission makers have as little resistance as possible when making missions, and this does raise the barrier; imo it raises it enough to be a blocker. When a mission maker decides to add an additional vehicle, they should not have to edit the vehicle properties at all to make it work with the loadouts.

addons/assignGear/functions/fnc_setBoxContentsFromConfig.sqf
TheCandianVendingMachine: I'm just going off of the blame. This is indeed not mine, viewing the history and blame pabst was the original author of the code path in `assignGearVehicle` which calls this. I'm just being a stickler about credits. Strictly speaking this should be credited to pabst as well, I must have just fucked up when I made my changes

addons/assignGear/functions/fnc_assignGearVehicle.sqf
TheCandianVendingMachine: Created this PR for ACE, hopefully it's merged. It will allow us to set the size to zero while still allowing some interaction time

https://github.com/acemod/ACE3/pull/10308

lambdatiger: I'd rather just increase the space, so all a mission maker would need to do is sum the number of boxes. I'll do that through the config like I did in #540 since I don't like opening the door to stacking innumerable boxes into one vehicle.

addons/assignGear/functions/fnc_setBoxContentsFromConfig.sqf
lambdatiger: Gotcha, at that point then this function is also just me (didn't notice I cut out the part I had originally credited you for)

addons/assignGear/functions/fnc_assignGearVehicle.sqf
lambdatiger: Done here https://github.com/BourbonWarfare/POTATO/pull/544/commits/5da6194249cd40441ba96f2aaa1c9307bd83dc60

TheCandianVendingMachine (approved):

lambdatiger:
Resolved merge conflcits

PR #546, opened 2024-09-14
Add CSW for GM MG3 tripod, Fagot, and MILAN.

lambdatiger:
This PR adds CSW version of GM static weapons including:
- MG3 AA tripod
- MILAN
- Fagot

To add the MILAN and Fagot, there is some abuse of the system by treating them as "tripods." The MG3 uses current ACE master functionality to allow primary weapons to be mounted. Video on discord.

### DIFF
--- a/addons/cswCompat/patchGM/config.cpp
+++ b/addons/cswCompat/patchGM/config.cpp
@@ -7,10 +7,17 @@
 class CfgPatches {
     class ADDON {
         units[] = {
-          "gm_ge_army_milan_launcher_tripod_csw"
+          "gm_ge_army_milan_launcher_tripod_csw",
+          "gm_gc_army_fagot_launcher_tripod_csw",
+          "gm_ge_army_mg3_aatripod_csw",
+          QGVAR(gm_MG3Tripod)
         };
         weapons[] = {
-          QGVAR(gm_milan)
+          QGVAR(gm_milan_backpack),
+          QGVAR(gm_fagot_backpack),
+          QGVAR(gm_MG3TripodCarry),
+          QGVAR(gm_milan_launcher_proxy),
+          QGVAR(gm_fagot_launcher_proxy)
         };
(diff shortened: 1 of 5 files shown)


addons/cswCompat/patchGM/config.cpp
```diff
@@ -0,0 +1,41 @@
...
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {
+            "potato_cswCompatCUP",
+            "gm_weapons_ammo",
+            "gm_weapons_launchers_milan",
+            "gm_weapons_launchers_fagot",
+            "gm_weapons_machineguns_mg3_gm_ge_army_mg3",
+            "gm_weapons_launchers_fagot_gm_gc_army_fagot"
+        };
+        skipWhenMissingDependencies = 0;
```
PabstMirror, line 31:
```suggestion
        skipWhenMissingDependencies = 1;
```

PabstMirror (approved):
looks good,
probably something we could just put in ace?

lambdatiger:
I abuse the tripod interface, since I don't think CSW supports weapons placed without them. So instead of placing a tripod and then a launcher, you place a tripod that *is* the launcher. Custom ACE action to pick it back up, so it's harder to inherit.

PabstMirror:
iirc that's how the dragon works

TheCandianVendingMachine:
> I abuse the tripod interface, since I don't think CSW supports weapons placed without them. So instead of placing a tripod and then a launcher, you place a tripod that _is_ the launcher. Custom ACE action to pick it back up, so it's harder to inherit. If that's ok, I can take time to make it into an actual ACE PR as well.

`dragon` as well as the compats for `metis` and `kornet` do this, its intended API

lambdatiger:
Alright, I'll put together an ACE PR at some point when I find the time to clean it up.

PR #550, opened 2024-09-18
Add custom version of Johnb43's Armor Modifier Ace

lambdatiger:
The mod adds a version of AMA
- Hit point armor minimum and maximum's set per side
- Unit specific hit point settings using `setVariable`
- API for setting both class default and specific unit armor values
- Some useful groupings of hit points such as "body", "limbs", "arms", and more (see fnc_resolveHitPoints)

Todo
- [x] Set defaults aimed at TvT damage
- [x] ~~Suppressed errors only for AMA~~ can't seem to pragma it away properly
- [x] MP testing (ty @TheCandianVendingMachine  & @AChesheireCat )

PR #551, opened 2024-09-19
Cleanup old settings and unused mg_redeye

PabstMirror:


TheCandianVendingMachine:
not unused
https://github.com/BourbonWarfare/POTATO/blob/master/addons/missileGuidanceCompat/patchCWR/config.cpp#L22

addons/settings/XEH_preInit.sqf
```diff
@@ -37,7 +37,7 @@ if (isServer) then {
         } forEach _settings;
 
         // report specific medical settings
-        private _log = format ["[AAA=%1] [aDmgPass=%2]", AAA_VAR_MOD_ENABLED, ace_medical_engine_damagePassThroughEffect toFixed 2];
+        private _log = format ["[aDmgPass=%1]", ace_medical_engine_damagePassThroughEffect toFixed 2];
```
lambdatiger, line 40: Already making a change here as part of #550. If you want it this way, let me know and I'll remove it from that PR to avoid the merge conflict.

PR #552, opened 2024-09-19
MG - Move redeye to cup

PabstMirror:


TheCandianVendingMachine (approved):
just for reference because I forgot my reasoning: the CUP Redeye reuses the `stinger` ammo class, despite being a different kind of ammo; hence the new defined class

PR #553, opened 2024-09-20
Settings - Update bw_settings for mod changes

PabstMirror:
- Settings changes for upcoming mod changes
- Store some settings that were set only on server

### DIFF
--- a/addons/settings/BW_Settings.inc.sqf
+++ b/addons/settings/BW_Settings.inc.sqf
@@ -142,8 +142,8 @@ if (["tuncon_unconinfo"] call ACEFUNC(common,isModLoaded)) then { _settings appe
 ["tuncon_unconinfo_enableshowheartrate", false],
 ["tuncon_unconinfo_enableshowstablevitals", false],
 ["tuncon_unconinfo_allownearestunit", false],
-["tuncon_unconinfo_nofriendliesnearbytext", "You are near death."]
-["tuncon_unconinfo_enableshowdetailedtreatment", false]
+["tuncon_unconinfo_nofriendliesnearbytext", "You are near death."],
+["tuncon_unconinfo_enableshowdetailedtreatment", false],
 ["tuncon_unconinfo_unconinfonearestunitdistance", 10]
 ];} else { _skip pushBack "tuncon_unconinfo" };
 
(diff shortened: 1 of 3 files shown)


addons/settings/BW_Settings.inc.sqf
```diff
@@ -42,6 +42,8 @@ _settings = [
 
 [QACEGVAR(advanced_fatigue,performanceFactor), 1.4],
 [QACEGVAR(advanced_fatigue,recoveryFactor), 1.6],
+[QACEGVAR(advanced_fatigue,terraingradientfactor), 0.4],
```
PabstMirror, line 45: this is new in ace 3.18, it makes walking on slopes and uneven terrain much more exhausting 
IIRC on first mission everyone was getting tired too quickly so we just changed it on the fly to this
I'm just copying what was saved

kil0byt3: was me, during that TVT we were walking directly uphill so i bumped it down pretty big, it'll be worth a re-look later, but these vals are fine for now

addons/settings/BW_Settings.inc.sqf
```diff
@@ -64,6 +66,7 @@ _settings = [
 [QACEGVAR(nightvision,aimDownSightsBlur), 0.25, false, true],
 [QACEGVAR(overheating,enabled), false],
 [QACEGVAR(refuel,rate), 10],
+[QACEGVAR(refuel,progressduration), 0.1],
```
kil0byt3, line 69: this might have been for testing, is this the time it takes to attach fuel line to vehicle?

kil0byt3 (approved):
rest looks good, we might want to server force `[QACEGVAR(finger,maxrange), 7],` as I have seen the mismatch pop up a lot for that one (missions are still often set to maxrange = 3 for some reason)?

addons/settings/BW_Settings.inc.sqf
PabstMirror: yeah, I think it was 0 on server,

kil0byt3: yea, should probably not be so low, but the default is slightly too long as well, can we make it like 2 seconds

PR #560, opened 2024-09-25
Fix AMA's Handling of Armor

lambdatiger:
We didn't dive super deep into how johnB was modifying the ACE damageHanlde funciton and that's on me for not double checking his work there.

He was modifying the variable `_realDamage`, which should be the damage calculated based on the bullet's speed/ `typicalSpeed` / `hit`. Changing this makes other medical functions it may be passed to treat it as if, more damaging projectiles were what hit the unit. `_newDamage` should be related to `_realDamage` by a factor of whatever armor is relevant in the ACE/AMA "handleDamage" EH function, which it isn't quite in the current version. 

If someone would like to do a simple test to double check I'd appreciate that.


lambdatiger:
Should be gtg, if there are any small features lmk. I'm also updating the defaults here so that future users will be left with AMA turned off on initialization.

PabstMirror (approved):

PR #562, opened 2024-09-26
adjusted some admin menu stuff

kil0byt3:
fixed some spelling and made some changes to the admin log list to make it wider (I think? I need to find a better way to edit GUIs)

this is a draft because 
- I'd like to test it 
- I'd like to make it wide enough to see most of the message without hovering
- I'd like to make the admin log stay on screen when the TP/healing menu is open, so you can reference which squad people are requesting TP to, etc

### DIFF
--- a/addons/adminComs/Interrupt_adminComs.hpp
+++ b/addons/adminComs/Interrupt_adminComs.hpp
@@ -35,8 +35,8 @@ class GVAR(textInput): RscEdit {
     font = "EtelkaMonospacePro";
 };
 class GVAR(sendMessageToAdmin): RscButtonMenu {
-    text = "MSG Admins";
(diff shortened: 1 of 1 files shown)


PabstMirror (approved):

PabstMirror:
image

overlaps a little, but both are still useable and I think it's fine

kil0byt3:
Looks a bit more messy but I like that we can see more of the message. 

Wondering if as an alternative, we just hide the timestamp in the list, and make the timestamp appear when you hover the message instead (as a tooltip) 

Then we could fit more of the message in the list

lambdatiger:
This still WIP?

PabstMirror:
I think it's an improvment for everyone so I don't see any reason not to merge
we're not going to get any more horizontal room without moving that code box which I really don't want to mess with

PR #566, opened 2024-10-08
Add an Auto Exit Safe Start Option for Mission Maker

lambdatiger:
This PR will add an option to force end Safe Start to the mission maker:

- A button to enable/disable force end safe start in the testing menu (off by default)
- Admin menu interface to delay the forced end to safe start if needed
  - UI elements to make it obvious from the admin menu whether the setting is set
- Added Mission Testing feedback
  - Side menu information
  - Testing question related to forcing safe start off

### DIFF
--- a/addons/adminMenu/Interrupt_adminMenu.hpp
+++ b/addons/adminMenu/Interrupt_adminMenu.hpp
@@ -1,3 +1,4 @@
+class ctrlCheckbox;
 class RscPicture;
 class RscFrame;
 class RscText;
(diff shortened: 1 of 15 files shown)


TheCandianVendingMachine (approved):
seems fine. nice feature, i like checkboxes being added to admin menu

lambdatiger:
Was brushing my teeth and realized there's a change to be made here. The current way it syncs over the network has race conditions that can cause client admins to desync their settings from the server. Instead the sync should throw a server event and then the server should put out a global event so changes queue properly.

lambdatiger:
Should be good now

PR #571, opened 2024-10-13
Remove DEBUG_MODE_FULL define from miscMedical postInit.sqf

lambdatiger:
As name says, every explosion generates at least 3 lines in rpt right now and I don't thing @TheCandianVendingMachine is currently reviewing it for anything.


lambdatiger:
The error is related to HEMTT v1.13.1 (which isn't up on winget yet) and is addressed in #569

PabstMirror (approved):

Issue #582, opened 2024-10-21
Add an ACE Painkiller "Blister Pack"


AChesheireCat:
Essentially just a lower use-count painkiller bottle - reference convo here: https://github.com/BourbonWarfare/bwmf/pull/203

kil0byt3:
https://discord.com/channels/204621032428929025/1293996766106222612/1297993522091393046

AChesheireCat:
I was JUST gonna comment that LMAOOOO

kil0byt3:
i just don't have time to do it rn but it could be added to bwmf common.hpp just via some use of that, probably, it wouldn't have a unique name or icon but it would work for your purposes

AChesheireCat:
If we're gonna do it, we should probably do it right imo. Plus, Naloxone was already added via POTATO so I'm inclined to keep similar stuff together instead of dispersing it across our various gits.

TheCandianVendingMachine:
that would need to be added to potato regardless. its a config update, so it has to be done mod side. you could technically script a pill bottle to start with less pills, but it would be hacky to do at the loadout level

PR #583, opened 2024-10-21
Adds a low-count painkiller alternative

AChesheireCat:
This is a first pass try - it is very much a WIP draft. I still need to test it to see if it works.

For context, reference the following:

https://github.com/BourbonWarfare/POTATO/issues/582

and

https://github.com/BourbonWarfare/bwmf/pull/203#issuecomment-2427566788

This is the aforementioned "blister pack" of painkillers.

- Only has 2 uses compared to the 10 that the bottle has

- Intent is to be issued in the basic kit so supplement medic capacity to treat pain


addons/miscMedical/ACE_Medical_Treatment_Actions.hpp
```diff
@@ -7,6 +7,8 @@ class ace_medical_treatment_actions {
         items[] = {"potato_naloxone"};
         litter[] = {{"ACE_MedicalLitter_atropine"}};
     };
+    class Painkillers: Morphine {};
+    class Pkblister: Painkillers {};
```
TheCandianVendingMachine, line 11: if we are assuming a blisterpack painkiller is the same as a painkiller, we dont need this

addons/miscMedical/Painkiller_Patch.hpp
```diff
@@ -0,0 +1,44 @@
+// License Attribution req: UI image authored by user grmarcstock via Vecteezy.com
+
+class CfgMagazines {
```
TheCandianVendingMachine, line 3: split into file `CfgMagazines.hpp` for consistency

addons/miscMedical/Painkiller_Patch.hpp
```diff
@@ -0,0 +1,44 @@
...
+    class potato_pkblister: ACE_painkillers {
+        author = "AChesheireCat";
+        displayName = "Painkiller Blister Pack";
+        descriptionShort = "For physical pain only";
+        descriptionUse = "A blister pack containing individualized painkiller tablets, regularly issued in IFAKs";
+        picture = QPATHOF(ui\pkblister_ca.paa);
+        count = 2;
+    };
+};
+class CfgVehicles {
```
TheCandianVendingMachine, line 15: split into file `CfgVehicles.hpp` for consistency

addons/miscMedical/Painkiller_Patch.hpp
```diff
@@ -0,0 +1,44 @@
...
+    class potato_pkblisterItem: ACE_painkillersItem {
+        author = "AChesheireCat";
+        displayName = "Painkiller Blister Pack";
+        vehicleClass = "Items";
+        class TransportItems {
+            MACRO_ADDMAGAZINE(potato_pkblister,1);
+        };
+    };
+};
+class CfgWeapons {
```
TheCandianVendingMachine, line 27: split into file `CfgWeapons.hpp` for consistency

addons/miscMedical/Painkiller_Patch.hpp
TheCandianVendingMachine, line 1: @lambdatiger does this fit in with `aceEdits`?

TheCandianVendingMachine (changes requested):
do we have the license to distribute the `pkblister_ca.paa`? aka, are we violating any copyright?

AChesheireCat:
> do we have the license to distribute the `pkblister_ca.paa`? aka, are we violating any copyright?
> 
> configs are all correct from what i can see. just some comments on the actual files

I think so - for reference this is the exact link for where I found the image file: https://www.vecteezy.com/vector-art/21385574-pill-tablet-design?autodl_token=608ca36d29a7922b511ffdf557b36320d9a8a5f3b6ac441f7a1ab75e1dd4eca108841db354e955ba929fd3f16f048f29f64c5792ac11f98b435c77da620b3fd7

The license listed on the page is royalty free with the only requirement being attribution by mentioning the author and the website, at least according to the licensing agreement that was included in the download. Honestly, if we don't want the headache I'm fine with just using the original pill bottle - this was just something cool to include lol.

TheCandianVendingMachine:
if we follow that license im all good with it, no headache at all

AChesheireCat:
HEMTT doesn't like having both Naloxone_Patch.hpp and the new broken-out files defining the same thing. Should I move the relevant sections in Naloxone_Patch.hpp (CfgVehicles, CfgWeapons) to the new files as well?

TheCandianVendingMachine:
yes, ideally. i dont think that we can compile with the two definitions regardless, so they need to be consolidated

addons/miscMedical/Painkiller_Patch.hpp
lambdatiger: It would have, had `aceEdits` existed before `miscMedical` the entire addons could be part of it. Now it's a better here as `miscMedical` has our other other treatment addition, thenaloxone injector.

I agree we should clean up how this addon is implemented (`CfgVehicles.hpp`/`CfgMagazines.hpp`/`CfgWeapons.hpp`)

addons/miscMedical/config.cpp
```diff
@@ -12,6 +12,28 @@ class CfgPatches {
...
+            timeTillMaxEffect = 1;
+            maxDose = 0.1;
+            onOverDose = QFUNC(Naloxone);
+            incompatibleMedication[] = {};
+            viscosityChange = 0;
+        };
+        class Painkillers;
+        class Pkblister: Painkillers {};
+    };
+};
```
lambdatiger, line 32: If we're spinning off all the configs into their own file, then `ace_medical_treatment` should live in a file called `ACE_Medical_Treatment.hpp`

lambdatiger (changes requested):

addons/miscMedical/config.cpp
AChesheireCat: Sounds good - wasn't sure where I should've put it, but this makes sense. Done!

addons/miscMedical/ACE_Medical_Treatment.hpp
```diff
@@ -0,0 +1,18 @@
...
+            hrIncreaseHigh[] = {0, 0};
+            timeInSystem = 180;
+            timeTillMaxEffect = 1;
+            maxDose = 0.1;
+            onOverDose = QFUNC(Naloxone);
+            incompatibleMedication[] = {};
+            viscosityChange = 0;
+        };
+        class Painkillers;
+        class Pkblister: Painkillers {};
```
PabstMirror, line 16:
```suggestion
        class Pkblister: Painkillers {
            items[] = {"potato_pkblister"};
        };
```

PabstMirror: err, we need something like this on the ace medcial actions?

lambdatiger (approved):
At some point we may want to make a model for the garbage left behind or the item itself, but the PR looks good and works as intended.

PabstMirror (approved):

PR #587, opened 2024-10-23
Revert RHS Change to ammo_Penetrator_Base

lambdatiger:
RHS adds config edits to create 30 submunitions to emulate spalling to the `ammo_Penetrator_Base`* base class that all vanilla and most modded penetrators are inherit from. This is a pretty cool effect and can be deadly, but also doesn't account for how some mods use penetrators. For example, CUP's 105mm M393A3 HEP round creates 8 penetrators, each which combined with RHS produce 30 spall submunitions.

This PR
- Removes that config edit from the `ammo_Penetrator_Base` class.
- Does not remove this effect from RHS penetrators

**If we want to still create these submunitions on non-RHS penetrators, that's fine. I can just tweak how some of the rounds create penetrators to account for it. Feedback welcome.**

See discord for videos. I usually enjoy this feature as we don't use any other form of spalling, but it looks like it causes some unintended issues.

\* Penetrators are spawned by HEAT and some AP rounds to emulate the penetration and damage to components the projectile does. This allows projectiles to have the capability to do more damage inside an armor, while also detonating outside a vehicle.


PabstMirror (approved):

lambdatiger:
For posterity, here's the [link to discord videos demonstrating this change](https://discord.com/channels/204621032428929025/285870164153597953/1298458430691999785)

PR #595, opened 2024-11-07
Briefing - Make ORBAT Updatable and Trigger a Update on Safe Start Ending

lambdatiger:
This PR does what the name says, it finds the orbat diary entry if a unit has it and replaces it, or adds one if a unit doesn't. This also allows the `uiSleep` time to be variable.


lambdatiger:
Anyone know if there's a reason to use `vehicle _x` vs `_x`? I changed it to the latter here and there must have been a reason at some point: https://github.com/BourbonWarfare/POTATO/pull/595/commits/29684ac6d91e23be0965274dc0cc73464d90064f

PabstMirror:
not sure, maybe a long time ago we used to put vic crews into their vics at the start
and the idea was it would show the tank for tank crews?

either way, the change is fine as it's more important to show the specific role

PabstMirror (approved):

PR #596, opened 2024-11-08
Spectate - Fix Missing Name on JIPs

lambdatiger:
This is a simple attempt to fix missing names on the JIP/respawned units in the spectate menu. It waits for an attempt at name update to take effect, under the assumption that eventually the name is properly synced.


PabstMirror:
could also do this on line 38
`if (_returnName != "AI: ") then { _unit setVariable [QGVAR(cachedName), _returnName]; };`
and only set the var if it's valid

PabstMirror (approved):

PR #601, opened 2024-11-13
Settings - Revert change to terraingradientfactor

PabstMirror:
ace changed the default to 0.1 from what it had been  https://github.com/acemod/ACE3/pull/10361/files
so we no longer need this, but only merge if we are going to update our ace copy


lambdatiger (approved):
Looks good, ty for fixing that in ACE as well

Issue #593, opened 2024-11-05
Kimi's HMD mod broke the vanilla "Twin Cannon 20mm" for vehicles with dynamic pylons


AChesheireCat:
Issue: The vanilla "Twin Cannon 20mm" dynamic pylon item, when equipped on a vehicle with dynamic pylons, does not appear in the weapon menu for either the pilot or the co-pilot despite being assigned it via the attributes pylon settings. This only occurs when Kimi's HMD mod is loaded.

Vehicles affected include but are not limited to:

- PLA Armored Pack's Z11WA (both versions)
- Vanilla WY-55 Hellcat (Renamed: AW159 Wildcat)
- Vanilla PO-30 Orca (Renamed: Ka-60 Kasatka)
- CUP CH-146 Griffon (dynamic loadout ver.)
- Embraer EMB 312 Tucano

Repro:

1.) Load the normal mod preset
2.) Go to the 3den Editor and put down any vehicle with a dynamic loadout and access to the "Twin Cannon 20mm"
3.) Play as the pilot/co-pilot and swap through the weapons available
4.) The "Twin Cannon 20mm" will not be available
5.) Exit the game and load all mods *without* the HMD mod (may have to unload the CUP and MELB compats due to dependencies)
6.) Re-do steps 2 and 3
7.)  "Twin Cannon 20mm" should now appear and be usable by pilot/co-pilot

AChesheireCat:
I don't know how the hell the HUD mod breaks this, but after doing a bunch of testing I found that the HMD mod was the culprit.

AChesheireCat:
First identified (at least publicly) here: https://discord.com/channels/204621032428929025/585642663202783236/1297637848404594789

kil0byt3:
bruh what the _fuck_ lol. Good find

probably a config overwrite somewhere? further investigation tomorrow

PR #603, opened 2024-11-14
Kimi's HMD 20mm Cannon Fix

AChesheireCat:
Kimi's HMD mod broke the inheritance for the "gatling_20mm" cannon, breaking it on any dynamic pylon vehicle.

I also updated the cursorAim, ballisticsComputer, and magazines to be in-line with other, similar weapons, and to include ACE ammunition.

I haven't taken a look at the Fixed Wing PBO to see if the same thing is happening, but this at least fixed it on the vehicles that I mentioned in https://github.com/BourbonWarfare/POTATO/issues/593#issue-2634620810. This specifically includes the Tucano as being able to use the 20mm cannons now.

### DIFF
--- a/addons/miscFixes/patchHMD/config.cpp
+++ b/addons/miscFixes/patchHMD/config.cpp
@@ -15,7 +15,7 @@ class CfgPatches {
     };
 };
 
-class cfgWeapons {
+class CfgWeapons {
     class Default;
     class CannonCore: Default {};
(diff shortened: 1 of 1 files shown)


lambdatiger (approved):
Looks like it fixes the issue described and doesn't fuck anything else down the inheritance stuff.

PabstMirror (approved):

PR #606, opened 2024-11-19
Cleanup some base classes

PabstMirror:


lambdatiger (approved):
Thanks for catching a few of my mistakes there, all of that looks good and works on my side.

Issue #592, opened 2024-10-27
Patch the CUP GTK Boxer HMG to have the M3M firerate


AChesheireCat:
IRL the GTK Boxer is equipped with the M3M instead of the M2 HMG as it is currently.

From wikipedia: 

> German vehicles are usually fitted with the FLW-200 from KMW, which can be fitted with either a 7.62 mm MG3 machine gun, a 12.7 mm M3M HMG or a 40 mm GMW automatic grenade launcher.

https://en.wikipedia.org/wiki/Boxer_(armoured_fighting_vehicle)#Design

lambdatiger:
Resolved by https://github.com/BourbonWarfare/POTATO/pull/611

PR #611, opened 2024-12-03
GTK Boxer Handling + Turret Tweaks

AChesheireCat:
https://github.com/BourbonWarfare/POTATO/issues/592

- Increased initial acceleration to make vehicle feel more responsive

- Increased braking torque to make braking feel more responsive

- Decreased fording depth due to being able to operate despite the engine and exhaust being submurged by +1m of water over the hull

- Swapped the gunner's HMG turret from the M2 to the M3M per the wikipedia page

- Moved the countermeasures smoke from the commander's turret to the gunner's turret

### DIFF
--- a/addons/miscFixes/patchCUP/config.cpp
+++ b/addons/miscFixes/patchCUP/config.cpp
@@ -56,6 +56,52 @@ class CfgVehicles {
             };
         };
     };
+    // Tweaks to the GTK Boxer's handling (accel/braking) + HMG swap to M3M + countermeasures move to gunner
+    class Wheeled_APC_F: Car_F {
+        class NewTurret;
+        class Turrets {
+            class MainTurret: NewTurret {
+                class Turrets {};
+            };
+        };
+    };
+    class CUP_Boxer_Base: Wheeled_APC_F {};
+    class CUP_Boxer_Base_HMG: CUP_Boxer_Base {
+        class CargoTurret;
+        class Turrets {
+            class MainTurret: NewTurret {
+                weapons[] = {"CUP_Vhmg_M3P_veh","SmokeLauncher"}; // was CUP_Vhmg_M2_veh
+                magazines[] = {"CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","SmokeLauncherMag"}; // was CUP_200Rnd_TE1_Red_Tracer_127x99_M
+            };
+            class CommanderTurret: NewTurret {
+                weapons[] = {}; // was "SmokeLauncher"
+                magazines[] = {}; // was "SmokeLauncherMag"
+            };
+        };
+        accelAidForceCoef = 2.3; // was 1.4
+        accelAidForceSpd = 25; // was 5
+        brakeIdleSpeed = 1.78; // was 0
+        maxFordingDepth = 1.0; // was 1.5
+        class Wheels {
+			class wheel_1_1 {
+				maxBrakeTorque = 20000; // was 12500
+				maxHandBrakeTorque = 30000; // was 25000
+			};
+		};
+    };
+    class CUP_Boxer_Base_GMG: CUP_Boxer_Base_HMG {
+        class CargoTurret;
+        class Turrets {
+            class MainTurret: NewTurret {
+                weapons[] = {"CUP_Vgmg_HKGMG_veh","SmokeLauncher"}; // added smoke launcher
+                magazines[] = {"CUP_32Rnd_40mm_MK19_M","CUP_32Rnd_40mm_MK19_M","CUP_32Rnd_40mm_MK19_M","CUP_32Rnd_40mm_MK19_M","CUP_32Rnd_40mm_MK19_M","CUP_32Rnd_40mm_MK19_M","SmokeLauncherMag"}; // added smoke mag
+            };
+            class CommanderTurret: NewTurret {
+                weapons[] = {}; // was "SmokeLauncher"
+                magazines[] = {}; // was "SmokeLauncherMag"
+            };
+        };
+    };
 };
 
 class CfgRecoils {
--- a/addons/miscFixes/patchCUP/config.cpp
+++ b/addons/miscFixes/patchCUP/config.cpp
@@ -65,7 +65,9 @@ class CfgVehicles {
             };
         };
     };
-    class CUP_Boxer_Base: Wheeled_APC_F {};
+    class CUP_Boxer_Base: Wheeled_APC_F {
+        class AnimationSources;
+    };
     class CUP_Boxer_Base_HMG: CUP_Boxer_Base {
         class CargoTurret;
         class Turrets {
@@ -88,6 +90,20 @@ class CfgVehicles {
 				maxHandBrakeTorque = 30000; // was 25000
 			};
 		};
+        class AnimationSources: AnimationSources {
+            class main_gun_muzzle_rot {
+                weapon = "CUP_Vhmg_M3P_veh";
+            };
+            class main_gun_reload {
+                weapon="CUP_Vhmg_M3P_veh";
+            };
+            class main_gun_reload_mag {
+                weapon="CUP_Vhmg_M3P_veh";
+            };
+            class main_gun_revolving {
+                weapon="CUP_Vhmg_M3P_veh";
+            };
+        };
     };
     class CUP_Boxer_Base_GMG: CUP_Boxer_Base_HMG {
         class CargoTurret;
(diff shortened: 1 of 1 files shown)


lambdatiger:
I'll test later tonight, but before I do, are the vehicle handling changes vibes based or something else? And is the 250 round mags accurate or for some other reason?

addons/miscFixes/patchCUP/config.cpp
```diff
@@ -56,6 +56,52 @@ class CfgVehicles {
...
+    // Tweaks to the GTK Boxer's handling (accel/braking) + HMG swap to M3M + countermeasures move to gunner
+    class Wheeled_APC_F: Car_F {
+        class NewTurret;
+        class Turrets {
+            class MainTurret: NewTurret {
+                class Turrets {};
+            };
+        };
+    };
+    class CUP_Boxer_Base: Wheeled_APC_F {};
```
lambdatiger, line 68: Not writing out the full sub classes in this class results in the classes in `CUP_Boxer_Bas_HMG` inheriting from the versions from `Wheeled_APC_F`. Luckily this time it doesn't cause issues since the `Turrets` class isn't changed in `CUP_Boxer_Base`, but it should be fixed to avoid any future issues.

 @PabstMirror lmk if I got something wrong above, but that's been my understanding

addons/miscFixes/patchCUP/config.cpp
```diff
@@ -56,6 +56,52 @@ class CfgVehicles {
...
+    class CUP_Boxer_Base_HMG: CUP_Boxer_Base {
+        class CargoTurret;
+        class Turrets {
+            class MainTurret: NewTurret {
+                weapons[] = {"CUP_Vhmg_M3P_veh","SmokeLauncher"}; // was CUP_Vhmg_M2_veh
+                magazines[] = {"CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","SmokeLauncherMag"}; // was CUP_200Rnd_TE1_Red_Tracer_127x99_M
```
lambdatiger, line 74: Just add the other 100 and 200 round boxes to the `CUP_Vhmg_M3P_veh` magazines array and stick with the 200 round boxes. That leaves most of the vehicle unchanged, while letting you edit magazine loadouts using the "potato" attributes tab.

addons/miscFixes/patchCUP/config.cpp
```diff
@@ -56,6 +56,52 @@ class CfgVehicles {
...
+        accelAidForceCoef = 2.3; // was 1.4
+        accelAidForceSpd = 25; // was 5
+        brakeIdleSpeed = 1.78; // was 0
+        maxFordingDepth = 1.0; // was 1.5
+        class Wheels {
+			class wheel_1_1 {
+				maxBrakeTorque = 20000; // was 12500
+				maxHandBrakeTorque = 30000; // was 25000
+			};
+		};
```
lambdatiger, line 90: I think the changes to breaking makes sense, but I'm not a huge fan of the acceleration changes. Acceleration feels weird to me (particularly like it's too high now), but that's with my current understanding that a Boxer essentially a very armored modular Stryker that somehow weight more than a Stryker when it doesn't have a module installed. I don't think it's blocking for the PR being merged, but something to highlight if anyone has objections.

lambdatiger (commented):
Works in game, but some minor feedback

addons/miscFixes/patchCUP/config.cpp
lambdatiger: Moving to a blocking review

addons/miscFixes/patchCUP/config.cpp
lambdatiger: Moving to a blocking review

addons/miscFixes/patchCUP/config.cpp
lambdatiger, line 68: Not writing out the full sub classes in this class results in the classes in CUP_Boxer_Bas_HMG inheriting from the versions from Wheeled_APC_F. Luckily this time it doesn't cause issues since the Turrets class isn't changed in CUP_Boxer_Base, but it should be fixed to avoid any future issues.

@PabstMirror lmk if I got something wrong above, but that's been my understanding

addons/miscFixes/patchCUP/config.cpp
lambdatiger, line 74: Just add the other 100 and 200 round boxes to the CUP_Vhmg_M3P_veh magazines array and stick with the 200 round boxes. That leaves most of the vehicle unchanged, while letting you edit magazine loadouts using the "potato" attributes tab.

lambdatiger (changes requested):

PabstMirror:
I have to look some more but there are some animations that are based on the weapon I think?
```
		class AnimationSources: AnimationSources
		{
		class main_gun_muzzle_rot
			{
				source = "ammorandom";
				weapon = "CUP_Vhmg_m2_veh";
			};

			class main_gun_reload
			{
				source="reload";
				weapon="CUP_Vhmg_m2_veh";
			};
			class main_gun_reload_mag
			{
				source="reloadmagazine";
				weapon="CUP_Vhmg_m2_veh";
			};
			class main_gun_revolving
			{
				source="revolving";
				weapon="CUP_Vhmg_m2_veh";
			};
```

lambdatiger:
> I have to look some more but there are some animations that are based on the weapon I think?

Good catch, I only double checked the particle effect memory points which should be good / compatible

addons/miscFixes/patchCUP/config.cpp
```diff
@@ -56,6 +56,68 @@ class CfgVehicles {
...
+                weapons[] = {"CUP_Vhmg_M3P_veh","SmokeLauncher"}; // was CUP_Vhmg_M2_veh
+                magazines[] = {"CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","CUP_250Rnd_TE1_Red_Tracer_127x99_M","SmokeLauncherMag"}; // was CUP_200Rnd_TE1_Red_Tracer_127x99_M
+            };
+            class CommanderTurret: NewTurret {
+                weapons[] = {}; // was "SmokeLauncher"
+                magazines[] = {}; // was "SmokeLauncherMag"
+            };
+        };
+        accelAidForceCoef = 2.3; // was 1.4
+        accelAidForceSpd = 25; // was 5
```
PabstMirror, line 84:
```suggestion
        accelAidForceCoef = 2.0; // was 1.4
        accelAidForceSpd = 15; // was 5
```
tone down accl changes?

addons/miscFixes/patchCUP/config.cpp
PabstMirror: I think it's ok as is, at least we don't seem to be getting any ubc

addons/miscFixes/patchCUP/config.cpp
PabstMirror: done, so loadouts are now like before exect for the smoke
also allows changing tracer color via potato_loadouts (m3p default is just red)

addons/miscFixes/patchCUP/config.cpp
lambdatiger: I can test it next year / maybe this weekend depending on some stuff goes, but if you think it doesn't feel out of place than you can merge it

lambdatiger (approved):

addons/miscFixes/patchCUP/config.cpp
lambdatiger: I thought that would be inheriting the previous class turret configs (`Wheeled_APC_F`), so if any other mod changes the `CUP_Boxer_Base`'s `turrets` class , we might override those changes even if we didn't change those specific parameters since we're changing the turret inheritance? Unless I misread the inheritance of the turret.

PabstMirror: CUP_Boxer_Base doesn't touch `turrets` as far as I can tell, only the hmg/gmg variants do

lambdatiger: Right, but if we load another mod that touches it `turrets` in CUP_Boxer_Base, wouldn't any change it makes be lost?

PabstMirror: the sub-classes fully redefine `turrets`
so any change to `CUP_Boxer_Base` would just be ignored

```
class CUP_Boxer_Base_HMG: CUP_Boxer_Base {
		class Turrets {
			class MainTurret: NewTurret { ... }
```

lambdatiger: Sounds good, ty.

addons/miscFixes/patchCUP/config.cpp
lambdatiger: Looks good

PR #612, opened 2024-12-04
add laser guided artillery

TheCandianVendingMachine:
does what it says. makes laser guided artillery actually be laser guided when not using arty computer


TheCandianVendingMachine:
using more esoteric cba macros purely because i was trying to fix an error that ended up not being related to macros at all, but then i had them so i kept them

lambdatiger (approved):
Looks good, works in game.

PabstMirror (approved):

PR #613, opened 2024-12-04
assingGear - Selectable "No optic" and Fixed List 

lambdatiger:
This PR adds a no optic option when custom optics are allowed and provides a CBA option to not hide the current optic from hte list.

### DIFF
--- a/addons/assignGear/XEH_preInit.sqf
+++ b/addons/assignGear/XEH_preInit.sqf
@@ -5,7 +5,6 @@ ADDON = false;
 PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
-#include "initSettings.inc.sqf"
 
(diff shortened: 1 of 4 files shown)


TheCandianVendingMachine (approved):
why `#` over select? operator precidence affecting this?

lambdatiger:
> why `#` over select? operator precidence affecting this?

I use # by default, do you want it to be select? Originally it was `primaryWeaponItems _player #2` which works but is less readable so I added the parenthesis.

TheCandianVendingMachine:
i dont care about it, just wanted to know if i had a blindspot

PabstMirror (approved):
lgtm
I don't think we need a setting or could make it default true but it's fine as is

lambdatiger:
I've been trying to avoid changing default behavior since I know some people like hiding the current optic. I think your modification helps make it a bit clearer which is current and we could get rid of it, let me do a quick "survey" and we can cut it out if there's no opposition.

lambdatiger:
Alright based on feedback I'll remove the option

PR #616, opened 2024-12-09
More vestBackpacks

AChesheireCat:
- Fixes some smersh vestBackpacks having incorrectly colored ui pictures

- Adds CWRIII ALICE LBE

- Adds M23 chest rigs in TKE and OD Green

- Adds 6Sh92 chest rigs

- Adds Raven vest

### DIFF
--- a/addons/customGear/README.md
+++ b/addons/customGear/README.md
@@ -4,6 +4,6 @@ A mod that serves as a place to put custom gear retextures, new gear items, and 
 
 ## Current Items
 
-- Converts some CUP vests (BAF Mk4 belt/webbing, RUAF Smersh in Olive and Coyote) to backpacks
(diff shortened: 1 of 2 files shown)


kil0byt3:
vackpacks

lambdatiger (commented):
All of your new (and old) backpacks converted from vests don't have the same container size and mass as the vests, which is ok. For example, the alice and alice mg vests both have a container space of 140 and mass of 15, while yours have maximum loads of 160 and 190, and masses of 25 and 30 respectively. I can't really identify your relationship between the two, could you help clarify?

AChesheireCat:
> I can't really identify your relationship between the two, could you help clarify?

Yeah, absolutely - these vest backpacks were initially intended to be used with vests that had limited-to-no carrying capacity so they needed to have a large enough carrying capacity to be able to fit our full combat load + whatever we typically put into our backpack's inventory (medical items, etc.). The weight is me thinking that it needed a balancing measure for how much they can carry if combined with vests with typical carrying capacities, but I'm so-so on that now.

lambdatiger:
Cool, I think you're the main consumer of the vest backpacks. So if it all works and you're happy with it, I'm happy with it.

lambdatiger (approved):

PabstMirror (approved):

PR #623, opened 2025-01-08
Zeus Utils - Initial and add Zeus FPS tracker

lambdatiger:
This adds a Zeus Utils addon and the initial version of a Zeus FPS tracker.

Tracker Features
- Mission enable (default off) CBA setting - Enabling the FPS collection/sending
- Client enable (default on) CBA setting  - whether the client should send it's FPS info
  - FPS is a moving average over the course of 5 seconds and can be sent every 3 to 90 seconds (default every 5)
  - Clients update server with their FPS
- Server sends FPS hash map to Zanes who have requested to receive updates.
  - Updates go out at a rate of 1/3 Hz (once every 3 seconds)
- Zeus can toggle receiving updates and has a few settings to configure display.
  - Zeus can decide if FPS above a threshold should be shown.
  - Zeus can configure low FPS threshold and low FPS font color with CBA settings


lambdatiger:
I'm unsure how well this scales, but I think because we can turn it off as it is, we should at least be able to turn it off if it's causing issues live.

Feature feedback also desired.

PabstMirror (approved):

PR #624, opened 2025-01-13
setVehicleAmmo - add setting for ace_rearm to reload the custom loadout

PabstMirror:
ref https://ace3.acemod.org/wiki/framework/rearm-framework#21-allow-rearming-of-scripted-loadouts


PabstMirror:
example of what this does:
ifv starts with red tracers
you change it's loadout to green tracers via potato
anytime you ace-rearm it would swap back to red, this should make ace refill it back to it's custom loadout

PR #626, opened 2025-01-21
Core - Add last group hint when player re-jips in

PabstMirror:
Adds hint for player's last group when the jip in to make it faster and easier to get them back to where they should be
Remove time from admin log and moves it to tooltip

### DIFF
--- a/addons/adminMenu/functions/fnc_uihook_teleportButton.sqf
+++ b/addons/adminMenu/functions/fnc_uihook_teleportButton.sqf
@@ -40,6 +40,13 @@ private _return = if (_vehicle == _leader) then {
         ["Teleporting to position", "No empty position for teleport"] select _success
     }
 } else {
+    [{ // hint if TP-to-Vic fails
+        params ["_selectedUnit"];
+        if (isNull objectParent _selectedUnit) then {
+            systemChat format ["%1 - failed to move into vic", name _selectedUnit];
+        };
+    }, [_selectedUnit], 2] call CBA_fnc_waitAndExecute;
+
     private _fullCrewCount = 0;
     private _aliveCrewCount = {
         _fullCrewCount = _fullCrewCount + 1;
--- a/addons/adminMenu/functions/fnc_uihook_teleportButton.sqf
+++ b/addons/adminMenu/functions/fnc_uihook_teleportButton.sqf
@@ -41,11 +41,11 @@ private _return = if (_vehicle == _leader) then {
     }
 } else {
     [{ // hint if TP-to-Vic fails
-        params ["_selectedUnit"];
-        if (isNull objectParent _selectedUnit) then {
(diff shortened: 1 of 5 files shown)


kil0byt3:
> Remove time from admin log and moves it to tooltip

amazing, both sound great

addons/adminMenu/functions/fnc_uihook_teleportButton.sqf
```diff
@@ -40,6 +40,13 @@ private _return = if (_vehicle == _leader) then {
         ["Teleporting to position", "No empty position for teleport"] select _success
     }
 } else {
+    [{ // hint if TP-to-Vic fails
```
TheCandianVendingMachine, line 43: (nitpick)
niche case, but if someone is fucking around in the start zone and enter a vehicle they won't have this pop

or if we add a shitty addon that for some reason makes this return an object

addons/core/functions/fnc_playerJipHint.sqf
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+
+if (isServer) then {
+    GVAR(playerMap) = createHashMap;
+    [QGVAR(playerConnected), {
+        params ["_player", "_profileName"];
+        private _msg = if (_profileName in GVAR(playerMap)) then {
+            (GVAR(playerMap) get _profileName) params ["_lastSide", "_lastGroup"];
+            private _sideText = ["?", "B", "O", "I"] select ([blufor, opfor, independent] find _lastSide);
```
TheCandianVendingMachine, line 14: civilian handling doesn't work, it will default to `I` since `find == -1`

addons/core/functions/fnc_playerJipHint.sqf
```diff
@@ -0,0 +1,40 @@
...
+            format ["has JIPed; last in [%1-%2]", _sideText, _lastGroup]
+        } else {
+            private _lastSide = side group _player;
+            private _lastGroup = groupId group _player;
+            GVAR(playerMap) set [_profileName, [_lastSide, _lastGroup]];
+            format ["has JIPed"];
+        };
+        INFO_2("playerConnected: %1: [%2]",_profileName,_msg);
+        if (CBA_missionTime < 60) exitWith {};
```
TheCandianVendingMachine, line 23: future:
can  probably set this timer to be the safe-start length as specified in the mission parameters option

TheCandianVendingMachine (changes requested):
imo the civilian case needs to be handled, we use them often enough for this to be annoying

addons/core/functions/fnc_playerJipHint.sqf
PabstMirror: opps, I think there should have been a subtraction here, I'll just swap to a `switch`

TheCandianVendingMachine (approved):
send it

PR #629, opened 2025-01-25
Misc Fixes - CUP A-10A CCIP Fix

lambdatiger:
This PR should make the A-10A HMD use CCIP calculation for Bombs,Guns, and Rockets


PabstMirror:
I know ccip is a sore-spot / meme for the cup people 
and they've had a long history of problems with it
so maybe give it a share with them as well

PabstMirror (approved):

lambdatiger:
> I know ccip is a sore-spot / meme for the cup people and they've had a long history of problems with it so maybe give it a share with them as well

I'll work through more of them and see if I can get an accurate HMD for all of them. I think at that point we could point Drofseh at the repo since last time I saw it sounded like he did part of them.

PR #631, opened 2025-01-28
Markers - Rewrite in Attempt to Provide Better Network Marker Sync and API for Admin Tools

lambdatiger:
This PR does a number of things
- Changes markers to rely on marker hash rather than initialization variables set on players
- Provides a method for markers to be kept in a networked, JIP compatible fashion
- Modifies ACE actions to
  - Make markers more immediately linked to the "become leader" self interact as well as 
  - Modify joining/leaving a group so you don't retain the marker
- Introduces four events to change markers
  - `addMarker` - For adding a new marker
  - `deleteMarker` - For deleting a marker
  - `transferMarker` - For transferring a marker within a group, transferring a unit marker, or detaching a marker
  - `updateMarker` - For changing the visual properties of a marker (text, icon, color, size)

This is still a WIP and will be ready for review when:
- [x] Integration with other addons (respawn, admin menu) are completed
- [x] Move from CBA events to `remoteExecCall` and other similar variable events for JIP queue
- [x] JIP-server marker position sync function is written (`publicVariableClient` an array of active marker positions to JIPs)
- [x] More thorough networked testing is completed
- [x] ~~Approval of inherently higher network overhead marker system is received~~
  - Someone has to review it either way ¯\\__(ツ)_/¯

### Any input is desired at this time, particularly on the technical aspects of this approach


lambdatiger:
This should be feature complete for what I think it should have, I'll work on doing some local testing aimed at:

- [ ] Client synchronization
  - [ ] Markers sync across all clients properly
  - [ ] Units can take, drop, or lose group markers and they sync properly
  - [ ] Dedicated server updates as expected
  - [ ] Respawns propagate properly
  - [ ] Ability to rebuild markers on the fly from the JIP CBA namespace object
- [ ] Handling update events from any client
  - [ ] Detach / reattach markers
  - [ ] Delete markers
  - [ ] Change marker visuals
  - [ ] JIP CBA namespace object should update/synchronize properly 
- [ ] JIP feature performance (not large scale performance testing, rather do features work)
  - [ ] Markers should initialize for mid session connections
  - [ ] Markers that have been dropped / aren't attached to an object or group should synchronize location

lambdatiger:
Thinking about this today, we may also be able to transfer from doing `drawIcon` to straight markers with 2.20 since it provides the ability to set marker color by RGB rather than fixed `CfgMarkerColors` colors. I'll put that in writing for now, but it would handle some of the more complex parts of this system (namely network syncing & JIP compatibility) fairly automatically.

lambdatiger:
This is good for review, if you have testing questions let me know!

lambdatiger:
> Thinking about this today, we may also be able to transfer from doing `drawIcon` to straight markers with 2.20 since it provides the ability to set marker color by RGB rather than fixed `CfgMarkerColors` colors. I'll put that in writing for now, but it would handle some of the more complex parts of this system (namely network syncing & JIP compatibility) fairly automatically.

RE this:
It'd be harder in some ways as well (side marks sometimes don't sync on join/respawn), unless every update was server side as well.

PabstMirror (approved):

PR #633, opened 2025-01-28
Admin Menu - Add markers Tab

lambdatiger:
When merged this PR will
- Add a markers tabs to the admin menu, both units with their respective markers as well as a list of all markers (all sides)
- Provide the ability to move a marker to a specific group or unit
- Ability to delete a marker
- Ability to detach a marker
- Ability to force reset all markers on all units

Before this PR is merged the following should occur:
- [x] ~~Add a refresh marker, for the off chance a unit changes variable name mid-mission~~
  - Shouldn't happen often if ever, primarily an issue with AI units 
- [x] Global reset marker event ~~(probably a public variable from the server to overwrite everyone's marker hash, but we'll see)~~
- [x] Return admin menu to only listing players
- [x] Merge #631


lambdatiger:
Useful comparison to see changes just to admin menu:
https://github.com/BourbonWarfare/POTATO/compare/Markers-Rewrite..markers-admin-menu

PabstMirror (approved):

PR #636, opened 2025-01-31
Misc Fixes - W35

lambdatiger:
W35 introduces a few changes to `B_Soldier_base_f`, some of which propagate to all (including BW) west units. The important one is the `accuracy` decrease making it harder for AI to spot players, but the other two are just things I would like reverted as well.

### DIFF
--- a/addons/miscFixes/patchW35/config.cpp
+++ b/addons/miscFixes/patchW35/config.cpp
@@ -1,6 +1,6 @@
 #include "\z\potato\addons\miscFixes\script_component.hpp"
 #undef COMPONENT
-#define COMPONENT miscFixes_patchWW35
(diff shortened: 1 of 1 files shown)


TheCandianVendingMachine:
a folder exists for these fixes here: https://github.com/BourbonWarfare/POTATO/blob/master/addons/aiCfgFixes/CfgVehicles.hpp

lambdatiger:
> a folder exists for these fixes here: https://github.com/BourbonWarfare/POTATO/blob/master/addons/aiCfgFixes/CfgVehicles.hpp. We might as well fix them for all time

Counter point, I don't want to create a one line compat in `aiCfgFixes` to make sure we load after W35

TheCandianVendingMachine (approved):

Issue #639, opened 2025-02-04
aiVehicleBail - Cookoff ACE Function Calls Do Not Function Properly


lambdatiger:
Cook off uses the pre-johnb432 overhaul function calls, some of which don't exist after Johnb432's rewrite. This results in cook-off, and specifically detonate events killing the crew without any sign that the vehicle was cooking off.

- As a note, some vehicles need their `potato_aiVehicleBail_<etc>Prob` updated/evaluated.

lambdatiger:
Per TCVM, should just be disabled, will instead move to cleanup or remove relevant settings and functions

TheCandianVendingMachine:
we can get rid of every function and event handler. should keep `CfgVehicles.hpp`, but thats just because I dont know if CUP has good VD configs

lambdatiger:
Yeah, already got it mostly done. Will do a double check of ACE vs our variables and then put the PR up for review.

PR #641, opened 2025-02-04
aiVehicleBial - Gut Functions and Config Update

lambdatiger:
This PR when merged will remove unused functions to avoid them being used addressing #639. In addition, it will fix Boxer hitpoint radius

- [x] ~~Review CUP ace_vehicle_damage config params and keep current params as needed.~~ We can keep it for now and do further review later


addons/settings/BW_Settings.inc.sqf
```diff
@@ -34,9 +34,6 @@ _settings = [
 
 // Vehicle Damage
 // [QEGVAR(aiVehicleBail,enableCrewBailing), false],
```
PabstMirror, line 36:
```suggestion
```

PabstMirror:
At some point I should try to figure out what values are different from base ace
and i guess see if we want to port them over or change ours
but it looks great for now

PabstMirror:
also weird that there is rhs stuff but not in reqAddons,
but should all be scope=1 so it doesn't really matter

PR #649, opened 2025-02-13
Markers - Fixes

lambdatiger:
This PR fixes various bugs with the recent markers update
- Attempt to fix JIP issues
- Mix missing SL post-fix
- Fix a rare reinforce issue when group spawned without FTL.

### DIFF
--- a/addons/markers/XEH_postInit.sqf
+++ b/addons/markers/XEH_postInit.sqf
@@ -36,6 +36,7 @@ LOG("Post init start");
 
                     [] call FUNC(checkForMapMarkerColor);
 
+                    // force reconstructino of drawHash
                     if (side (group _newPlayer) != side (group _oldPlayer)) then {
                         GVAR(drawHash) = createHashMap;
                         GVAR(nextUpdateDrawHash) = -1;
@@ -43,15 +44,18 @@ LOG("Post init start");
                     // not a fan of waiting here, is there a better way?
                     // the respawn setVariable seems to happen/propagate after this event
                     [{
-                        _this getVariable [QGVAR(addMarker), false] ||
-                        (group _this) getVariable [QGVAR(addMarker), false]
-                    }, {[_this] call FUNC(initUnitMarkers);},_newplayer, 10, {
-                        [_this] call FUNC(initUnitMarkers);
+                        (_this#0) getVariable [QGVAR(addMarker), false] ||
+                        (group (_this#0)) getVariable [QGVAR(addMarker), false]
+                    }, {_this call FUNC(initUnitMarkers);},[_newplayer], 10, {
+                        _this call FUNC(initUnitMarkers);
                     }] call CBA_fnc_waitUntilAndExecute;
                 }] call CBA_fnc_addPlayerEventHandler;
                 if (didJIP) then {
                     [true] call FUNC(reinitMarkerHash);
                 };
+                // In an attempt to catch those who join in map screen and
+                // miss the reinitMarkerHash call we call it after mission start
+                // without the network sync
                 [{
                     if (GVAR(drawHash) isEqualTo createHashMap) then {
                         call FUNC(reinitMarkerHash);
--- a/addons/markers/XEH_postInit.sqf
+++ b/addons/markers/XEH_postInit.sqf
@@ -45,15 +45,21 @@ LOG("Post init start");
                     [{
                         _this getVariable [QGVAR(addMarker), false] ||
                         (group _this) getVariable [QGVAR(addMarker), false]
-                    }, {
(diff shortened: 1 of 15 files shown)


addons/markers/XEH_postInit.sqf
```diff
@@ -36,25 +36,34 @@ LOG("Post init start");
...
                 }] call CBA_fnc_addPlayerEventHandler;
                 if (didJIP) then {
                     [true] call FUNC(reinitMarkerHash);
                 };
-            };
+                // In an attempt to catch those who join in map screen and
+                // miss the reinitMarkerHash call we call it after mission start
+                // without the network sync
+                [{
+                    if (GVAR(drawHash) isEqualTo createHashMap) then {
```
TheCandianVendingMachine, line 60: god i hate this syntax to refer to a hashmap. rare dedmen L

addons/respawn/CfgRespawnGroups.hpp
```diff
@@ -118,33 +118,43 @@ class CfgRespawnGroups {
         class Configurations {
             class ASL: Alpha {
                 displayName = "Alpha";
+                markerText = "SL";
```
TheCandianVendingMachine, line 121: does this fix the E bug you ran into?

TheCandianVendingMachine (approved):
not my favourite solution for the JIP problem but I can't think of anything better

addons/respawn/CfgRespawnGroups.hpp
lambdatiger: yes

addons/markers/XEH_postInit.sqf
lambdatiger: Same

lambdatiger: There's no flag for was there at mission start but before entered mission from map screen. Maybe we have the server set a variable on everyone connected at mission start and make our own?

TheCandianVendingMachine: that would probably be best, i did this for a JIP thing once and it seemed to be possible. imo different PR though

lambdatiger: lmk if you still have that somewhere, or if it's something we could hook into rather than having multiple systems do the same thing.

TheCandianVendingMachine: sadly i dont :( it was for another group when people asked me

lambdatiger:
Delta:
https://github.com/BourbonWarfare/POTATO/compare/7d142ca..abc1f1d

addons/markers/XEH_postInit.sqf
```diff
@@ -36,30 +36,39 @@ LOG("Post init start");
...
-                    }, {
-                        [_this] call FUNC(initUnitMarkers);
-                    }, _newplayer, 5
-                    ] call CBA_fnc_waitUntilAndExecute;
+                        (_this#0) getVariable [QGVAR(addMarker), false] ||
+                        (group (_this#0)) getVariable [QGVAR(addMarker), false]
+                    }, {_this call FUNC(initUnitMarkers);},[_newplayer], 10, {
+                        _this call FUNC(initUnitMarkers);
+                    }] call CBA_fnc_waitUntilAndExecute;
                 }] call CBA_fnc_addPlayerEventHandler;
```
PabstMirror, line 52:
```suggestion
                }] call CBA_fnc_addPlayerEventHandler; // no-retro flag set, this will only run for units that respawn
```

lambdatiger: yeah, that's on purpose. Good clarifying comment

PabstMirror (approved):

addons/markers/XEH_postInit.sqf
```diff
@@ -36,30 +36,39 @@ LOG("Post init start");
...
-                        (group _this) getVariable [QGVAR(addMarker), false]
-                    }, {
-                        [_this] call FUNC(initUnitMarkers);
-                    }, _newplayer, 5
-                    ] call CBA_fnc_waitUntilAndExecute;
-                }] call CBA_fnc_addPlayerEventHandler;
+                        (_this#0) getVariable [QGVAR(addMarker), false] ||
+                        (group (_this#0)) getVariable [QGVAR(addMarker), false]
+                    }, {_this call FUNC(initUnitMarkers);},[_newplayer], 10, {
+                        _this call FUNC(initUnitMarkers);
```
PabstMirror, line 50: really doesn't matter for this but newer hemtt will complain
```suggestion
                    }, {call FUNC(initUnitMarkers);},[_newplayer], 10, {
                        call FUNC(initUnitMarkers);
```

lambdatiger: If it works, it works. Happy to do the change

PR #653, opened 2025-02-20
missile guidance - re-enable mclos

PabstMirror:
ref https://github.com/acemod/ACE3/pull/10701
Very basic. but will present some threat if not moving

### DIFF
--- a/addons/missileGuidanceCompat/patchCUP/config.cpp
+++ b/addons/missileGuidanceCompat/patchCUP/config.cpp
@@ -18,6 +18,16 @@ class CfgPatches {
 #include "../CfgMissileTypesWarsaw.hpp"
 #include "../CfgMissileTypesNato.hpp"
 
+class CfgWeapons {
+    class missiles_SCALPEL;
+    class CUP_Vmlauncher_AT2_veh: missiles_SCALPEL {
+        ACEGVAR(missileGuidance,hasMCLOSControl) = 1;
+    };
+    class CUP_Vmlauncher_AT3_veh: missiles_SCALPEL {
+        ACEGVAR(missileGuidance,hasMCLOSControl) = 1;
+    };
+};
+
 class CfgMagazines {
     class CA_LauncherMagazine;
     class CUP_Dragon_EP1_M: CA_LauncherMagazine {
--- a/addons/missileGuidanceCompat/patchCUP/config.cpp
+++ b/addons/missileGuidanceCompat/patchCUP/config.cpp
@@ -18,16 +18,6 @@ class CfgPatches {
 #include "../CfgMissileTypesWarsaw.hpp"
 #include "../CfgMissileTypesNato.hpp"
 
-class CfgWeapons {
-    class missiles_SCALPEL;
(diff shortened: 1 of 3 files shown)


addons/missileGuidanceCompat/patchGM/config.cpp
```diff
@@ -100,24 +100,24 @@ class CfgAmmo {
             enabled = 1;
             showTrail = 1;
         };
-    };*/
+    };
     // Milan
-   /* class gm_missile_milan_base: gm_missile_saclos_base {
+   class gm_missile_milan_base: gm_missile_saclos_base {
```
lambdatiger, line 105: MILAN is SACLOS and was disabled because the turret-on-turret implementation of MG didn't get the right aim vector and would send the missiles off course unless the turret and body were facing 12 o'clock.

addons/missileGuidanceCompat/patchCUP/config.cpp
```diff
@@ -48,8 +48,8 @@ class CfgAmmo {
     };
 
     class CUP_M_9M17P_AT3_Sagger_AT: MissileBase {
-        //maneuvrability = 0;
-        //ACE_MISSILE(Fleyta);
+        maneuvrability = 0;
+        ACE_MISSILE(Fleyta);
```
lambdatiger, line 52: 9M17P is SACLOS, so the ACE simulation wouldn't be accurate for this

addons/missileGuidanceCompat/patchGM/config.cpp
```diff
@@ -85,7 +85,7 @@ class CfgAmmo {
         };
     };
     // Malyutka
-    /*class gm_missile_maljutka_base: gm_missile_saclos_base {
+    class gm_missile_maljutka_base: gm_missile_saclos_base {
```
lambdatiger, line 88: I think this is fine, AI seems to be able to aim, but I definitely have some control issues / can't seem to control the missile on my side.

addons/missileGuidanceCompat/patchCUP/config.cpp
lambdatiger: Bailey would like a new magazine, we can do this in a different PR, or if it's still up tomorrow I'll get around to it.

addons/missileGuidanceCompat/patchGM/config.cpp
lambdatiger: Would someone be able to check to make sure MCLOS works before I approve this? I'm not sure if it's not working or a skill issue on my part.

lambdatiger: Does this mean you'd need a separate weapon to fire a non-MCLOS missile?

lambdatiger (approved):

addons/missileGuidanceCompat/patchGM/config.cpp
PabstMirror: yeah these make it a pain, ace changed to not require weapon configs anymore

lambdatiger (approved):
Still looks good & works with latest ACE changes

kil0byt3 (approved):

PR #654, opened 2025-02-23
vehicleCoasting - Introduce a system for vehicles to continue driving after driver death or other catasrophic event

lambdatiger:
This PR introduces a system that allows vehicles (AI or player) to continue driving after driver is killed or other catastrophic event that would result in the vehicle escaping control of anyone alive. Currently events include:
- The driver dying
- Vehicle cookoff
- Vehicle engine cookoff

**TODO**
- [ ] Look at changing from "EntityKilled" missionEventHandler to another method
- [ ] Test it more until it feels perfect


addons/vehicleCoasting/XEH_PREP.hpp
```diff
@@ -0,0 +1,10 @@
+PREP(addCoastingVehicle);
+PREP(addVehicleHandle);
```
PabstMirror, line 2: Script \z\potato\addons\vehiclecoasting\functions\fnc_addVehicleHandle.sqf not found
Script \z\potato\addons\vehiclecoasting\functions\fnc_dev_trackVehicle.sqf not found
Script \z\potato\addons\vehiclecoasting\functions\fnc_driverDeathPFEH.sqf not found

addons/vehicleCoasting/functions/fnc_vehicleCookOffHandle.sqf
```diff
@@ -0,0 +1,44 @@
...
+* [_unit] call potato_vehicleCoasting_fnc_driverDeathHandle;
+*//***************************************************************************/
+params [["_vehicle", objNull, [objNull]]];
+TRACE_2("params",_vehicle,typeOf _vehicle);
+
+if !(alive _vehicle && local _vehicle &&
+    !((getObjectID _vehicle) in GVAR(activeVehicles))) exitWith {
+    TRACE_3("invalid vehicle",alive _vehicle,local _vehicle,alive _driver);
+};
+#define DEBUG_MODE_FULL
```
PabstMirror, line 26:
```suggestion
#ifdef DEBUG_MODE_FULL
```

addons/vehicleCoasting/functions/fnc_driverDeathHandle.sqf
```diff
@@ -0,0 +1,49 @@
...
+    {!(_unit isKindOf "CAManBase")}) exitWith {
+    TRACE_2("Invalid",isNull objectParent _unit,_unit isKindOf "CAManBase");
+};
+
+private _vehicle = vehicle _unit;
+
+if (!local _vehicle) exitWith {
+    TRACE_1("Not local",_vehicle);
+};
+#define DEBUG_MODE_FULL
```
PabstMirror, line 31:
```suggestion
#ifdef DEBUG_MODE_FULL
```

addons/vehicleCoasting/functions/fnc_vehicleCookOffHandle.sqf
```diff
@@ -0,0 +1,44 @@
...
+*
+* Example:
+* [_unit] call potato_vehicleCoasting_fnc_driverDeathHandle;
+*//***************************************************************************/
+params [["_vehicle", objNull, [objNull]]];
+TRACE_2("params",_vehicle,typeOf _vehicle);
+
+if !(alive _vehicle && local _vehicle &&
+    !((getObjectID _vehicle) in GVAR(activeVehicles))) exitWith {
+    TRACE_3("invalid vehicle",alive _vehicle,local _vehicle,alive _driver);
```
PabstMirror, line 24:
```suggestion
    TRACE_3("invalid vehicle",alive _vehicle,local _vehicle,alive driver _vehicle);
```

addons/vehicleCoasting/functions/fnc_addCoastingVehicle.sqf
```diff
@@ -0,0 +1,62 @@
...
+
+_vehicle setEffectiveCommander _driver;
+_vehicle disableBrakes true;
+
+private _freeRunLen = 20 + random 40; // Max runtime
+// Simulate broken tracks running out
+if (_vehicle getHitPointDamage "hitrtrack" >  0.8 ||
+        {_vehicle getHitPointDamage "hitltrack" >  0.8}) then {
+    _vehicle setHitPointDamage ["hitrtrack", 0.4, false];
+    _vehicle setHitPointDamage ["hitltrack", 0.4, false];
```
PabstMirror, line 34: I'm not sure?
```suggestion
    _vehicle setHitPointDamage ["hitrtrack", 0.4 min (_vehicle getHitPointDamage "hitrtrack"), false];
    _vehicle setHitPointDamage ["hitltrack", 0.4 min (_vehicle getHitPointDamage "hitltrack"), false];
```

PabstMirror (approved):

addons/vehicleCoasting/functions/fnc_addCoastingVehicle.sqf
lambdatiger: That makes sense since the current approach forces damage on undamaged tracks

lambdatiger:
Actually issue with that last one. I was about to merge another commit so give me a second, but there are issues with how/when those checks overlap

lambdatiger:
Ok so simple solution is to just apply to players, but the issue is maybe ~50% of the time we actually knock AI uncon (triggering the uncon event) and then they are killed quietly by ACE, triggering a death event. Since we don't play with AI uncon, I don't think it matters much here, but the other option would be to check that AI uncon is enables and use that instead.

lambdatiger:
Since we can disable it, I'm ok with us taking it for now, just setting the recommended percentage to a 40% instead of 50%.

PabstMirror (approved):

PR #655, opened 2025-02-23
zeusUtils - Add Draw Markers Button

lambdatiger:
This PR adds the ability to draw markers above the marker's holder. 

- It does not show markers for dead units or for markers detached for units
- The marker size is scale with distance and to marker map size
- Custom local settings for:
  - Max marker draw distance
  - Marker size
  - Displaying marker text
- Created common Draw3D EH
- Updated button UI to be more readable

### DIFF
--- a/addons/zeusUtils/CfgUI.hpp
+++ b/addons/zeusUtils/CfgUI.hpp
@@ -5,13 +5,22 @@ class RscDisplayCurator {
             idc = ZEUSUTILS_IDC_DISPLAYFPSBUTTON;
             // based on Potato's ACRE spectate button
             x = "safezoneX";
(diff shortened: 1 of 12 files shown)


PabstMirror:
I'm just wondering about putting the marker stuff for spectate as well?

lambdatiger:
> I'm just wondering about putting the marker stuff for spectate as well?

We definitely could. I would want to do it in a slightly different way since this implementation doesn't draw dropped/dead markers (on purpose / by request) and I think that could be pretty neat to see in spectate. Definitely could merge the controls between the two as well.

PabstMirror (approved):

PR #663, opened 2025-03-02
miscMedical - Estimate explosive filler via `ace_frag_gurney_c`

TheCandianVendingMachine:
title. should better estimate explosives, and probably be more accurate


lambdatiger:
I'd rather it just picked the proper values and defaulted to TNT if it's not present. I don't think there's a super strong correlation between gurney constant and TNT equivalence:
image
Line of best fit has an $$R^2$$ of ~0.015.

Let me know if I missed something

TheCandianVendingMachine:
the idea was mostly to estimate the filler if a mod does something custom/is off by a small value. in general the numbers should be within a few units of the filler; thats why the default is comp-b. if we want i guess we can set a window the difference must be in to be considered

lambdatiger (approved):
I'm a larger fan of a LUT using hashmaps, but this works as well assuming values are close.

PR #667, opened 2025-03-09
assignGear - Fix Supply Boxes Config Skipping

lambdatiger:
Sometimes `fnc_assignGearSupplyBox` skips child config classes. It would seem `configProperties` doesn't always work in `missionConfigFile`.


PabstMirror (approved):

PabstMirror:
Was it something like 
```
class car {
  class a {...};
};
class red_car: car {
  class b {};
}
```
I guess configClasses would only show B for red_car?

lambdatiger:
> Was it something like
> 
> ```
> class car {
>   class a {...};
> };
> class red_car: car {
>   class b {};
> }
> ```
> 
> I guess configClasses would only show B for red_car?

Yeah, particularly inherited classes.

PR #670, opened 2025-03-11
zeusUtils - Fix distance position type

### LINKED ISSUE #669: zeusUtils - Group Markers
Current implementation compares ASL position of units to position of camera object (expects AGL), making Zeus group markers look small.

PabstMirror:
Fix #669

### DIFF
--- a/addons/spectate/functions/fnc_setup.sqf
+++ b/addons/spectate/functions/fnc_setup.sqf
@@ -117,9 +117,9 @@ GVAR(hasTarget) = true;
 GVAR(surfaceSpeed) = false;
 GVAR(camTarget) = _oldUnit;
 GVAR(targetInVehicle) = !isNull objectParent GVAR(camTarget);
-GVAR(dummy) = "Logic" createVehicleLocal getPosASLVisual GVAR(camTarget);
+GVAR(dummy) = "Logic" createVehicleLocal ASLToAGL getPosASLVisual GVAR(camTarget);
 
-GVAR(cam) = "camcurator" camCreate eyePos GVAR(camTarget);
+GVAR(cam) = "camcurator" camCreate ASLToAGL eyePos GVAR(camTarget);
 GVAR(cam) cameraEffect ["internal", "back"];
 GVAR(cam) setPosASL eyePos GVAR(camTarget);
 GVAR(cam) setDir getDirVisual GVAR(camTarget);
--- a/addons/spectate/functions/fnc_ui_handleKeyDown.sqf
+++ b/addons/spectate/functions/fnc_ui_handleKeyDown.sqf
@@ -79,7 +79,7 @@ if ((inputAction "CuratorInterface") > 0 && {!isNull (getAssignedCuratorLogic pl
     [
         {!isNull curatorCamera},
         {
-            curatorCamera setPosASL positionCameraToWorld [0,0,0];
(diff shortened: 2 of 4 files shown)


addons/spectate/functions/fnc_setup.sqf
```diff
@@ -117,9 +117,9 @@ GVAR(hasTarget) = true;
 GVAR(surfaceSpeed) = false;
 GVAR(camTarget) = _oldUnit;
 GVAR(targetInVehicle) = !isNull objectParent GVAR(camTarget);
-GVAR(dummy) = "Logic" createVehicleLocal getPosASLVisual GVAR(camTarget);
+GVAR(dummy) = "Logic" createVehicleLocal ASLToAGL getPosASLVisual GVAR(camTarget);
 
-GVAR(cam) = "camcurator" camCreate eyePos GVAR(camTarget);
+GVAR(cam) = "camcurator" camCreate ASLToAGL eyePos GVAR(camTarget);
```
lambdatiger, line 122: Not sure if this is needed since it's `setPosASL`'d 2 lines down.

addons/spectate/functions/fnc_ui_handleKeyDown.sqf
```diff
@@ -93,13 +93,13 @@ if (inputAction "headlights" > 0) exitWith {
         { deleteVehicle _x; } forEach GVAR(camLights);
         GVAR(camLights) = [];
     } else {
-        private _cameraLight = "#lightpoint" createVehicleLocal getPosASL GVAR(cam);
+        private _cameraLight = "#lightpoint" createVehicleLocal ASLToAGL getPosASL GVAR(cam);
```
lambdatiger, line 96: Same as before, we use `attachTo` with an offset so the position shouldn't matter unless I'm missing something. Shouldn't be an issue here since it happen infrequently.

addons/zeusUtils/functions/fnc_drawMarkers.sqf
```diff
@@ -30,7 +30,7 @@ private _showMarkerText = GVAR(showMarkerText);
     if (isNull _drawObject) then {continue};
 
     private _baseDrawPos = (getPosASLVisual _drawObject) vectorAdd [0, 0, 15];
-    private _distance = _baseDrawPos distance _localCuratorModule;
+    private _distance = (ASLToAGL _baseDrawPos) distance _localCuratorModule;
```
lambdatiger, line 33: Solution I had (didn't PR it last night) was to store the the ASL position of the position of `_localCuratorModule` in a variable before the `forEach` loop and use that for distance calculation instead. Slightly more efficient, especially when there are 30-40 markers on the screen at once.

lambdatiger (changes requested):
There's on optimization I would suggest (last comment), but the first two are just thoughts if you don't change/I'm wrong about changing them let me know, but overall it it shouldn't matter since they happen infrequently.

addons/spectate/functions/fnc_setup.sqf
PabstMirror: I was just fixing anything my linter saw to make the warnings go away but yeah this really is pointless

addons/zeusUtils/functions/fnc_drawMarkers.sqf
PabstMirror: that sounds great, I'll just close this for now if you want to give it a go

Issue #413, opened 2023-07-04
Webknight's double bind shoulder weapon


kil0byt3:
"its default same as the ace one, making neither work."

This issue looks like it was mistakenly posted to [bwmf](https://github.com/BourbonWarfare/bwmf/issues/179)

I'm not sure we can ship default keybinds in our bat file, so not sure anything can be done but just moving the issue to the correct repo

kil0byt3:
not sure anything can be done other than making people aware. 

also i think it does work by default now (`0` puts weapon away without need to change keybinds)

PR #672, opened 2025-03-18
MiscFixes - Clear bad melee: hide weapon keybind

PabstMirror:


lambdatiger:
Could we run a quick poll to make sure no one uses it at all? Just in case we're screwing over a few members.

PabstMirror:
I'll just add that in my quick testing their keybind seemed to work fine
maybe they fixed it?

but I just remember this being requested

lambdatiger:
Definitely was, I think it's a nice QoL thing, especially for recruits.

lambdatiger (approved):
Looks good

kil0byt3:
Solves #413

PabstMirror:
I don't want to merge this unless we know the WK keybind isn't working

PR #680, opened 2025-04-04
Debriefing - Add countdown timer

PabstMirror:


lambdatiger:
Maybe change [the style](https://community.bistudio.com/wiki/Arma:_GUI_Configuration#Control_Styles) to `style = "0x02";`, so the text isn't uppercase (currently `style = "0x02 + 0xC0";`).

lambdatiger (approved):
Looks good to me, works well and should help slow down logged admin progression through the debrief.

PR #681, opened 2025-04-05
miscFixes - Adds Compat for CWR3 BAF 

lambdatiger:
This PR

- Adds a FV107 with realistic turret and weapon
- Adds MG for CWR3 BAF Javelin SAM

### DIFF
--- a/addons/missileGuidanceCompat/patchCWRBAF/config.cpp
+++ b/addons/missileGuidanceCompat/patchCWRBAF/config.cpp
@@ -1,0 +1,75 @@
+#include "\z\potato\addons\missileGuidanceCompat\script_component.hpp"
+#undef COMPONENT
+#define COMPONENT missileGuidanceCompat_patchCWRBAF
+
+class CfgPatches {
+    class ADDON {
+        units[] = {};
+        weapons[] = {};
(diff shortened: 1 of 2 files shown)


addons/missileGuidanceCompat/patchCWRBAF/config.cpp
```diff
@@ -0,0 +1,75 @@
...
+        VERSION_CONFIG;
+    };
+};
+
+class CfgAmmo {
+    class MissileBase;
+    class CUP_M_Stinger_AA: MissileBase {
+        class ace_missileguidance;
+    };
+    class GVAR(M_javelin_AA): CUP_M_Stinger_AA {
```
TheCandianVendingMachine, line 23: this looks mostly copies from the RBS-70; any reason why not just inheriting and changing values? (`class ace_missileguidance_rbs70`)

lambdatiger: That was a lack of cleanup on my part. I was originally going to write another navigation profile for it, but I ended up forgetting about it a month ago and thought I should get what I have out.

PabstMirror (approved):

PR #685, opened 2025-04-12
aceEdits - Add GBU-39B/B Laser SDB

lambdatiger:
## This PR adds:

- A 4x and 1x LSDB pylon and 1x SDB pylon
- A multi seeker, seeker type that stacks a number of seekers together (works well for SALH + GPS, may not be performant for others)
  - Prioritizes SALH over GPS.
- Add mass and ACE frag params for updated SDB

## Known bugs
- ~~Because AI  does not generate GPS target data on firing, if they drop on laser targets that are later lost the projectile moves towards world origin.~~


TheCandianVendingMachine (approved):
idk if `aceEdits` is the right place, but i dont think its worth changing

lambdatiger:
> idk if `aceEdits` is the right place, but i dont think its worth changing

Where would you put it? It felt weird to add functions to `missileGuidanceCompats` and `miscFixes` doesn't need more sub-addons.

TheCandianVendingMachine:
`missileGuidanceCompats` makes the most sense imo, but putting functions in complicates it. keep it here, its just a hard one to place

lambdatiger:
Sorry about round two, checked today and I made some mistakes

PabstMirror (approved):

PR #686, opened 2025-04-13
Admin Menu - Add guide to markers

PabstMirror:
text from LT's post


lambdatiger (approved):
The second one is a bit redundant as you can hover over each button in the menu, but it's fine. I added that to the explanation because it wasn't interactive.

PabstMirror:
I think there is a limit of ~20 lines, so I think all these should fit just fine

PR #693, opened 2025-04-26
CUP Helicopter Retextures

AChesheireCat:
Adds the following retextures for the CUP UH-60M and CH-47F helicopters.

UH-60M:
- Australian
- Askeland Defense Force

CH-47F:
- Australian
- Australian (Gray)


addons/miscFixes/patchCUP/config.cpp
```diff
@@ -196,6 +196,59 @@ class CfgVehicles {
     class CUP_Tigr_M_KORD_Base: CUP_Tigr_Base {
         CUP_MAINTURRET_DISABLE_ATTENUATE;
     };
+    
+    // CUP UH-60M
+    class Helicopter; // might need to inherit from Air; per the CH-47 config?
+    class Helicopter_Base_F: Helicopter {};
+    class Helicopter_Base_H: Helicopter_Base_F {};
```
lambdatiger, line 203: We shouldn't need to include the entire inheritance here.
```suggestion
    class Helicopter_Base_H;
```

lambdatiger (commented):
Do we need to do anything with licenses by including this?
ADF RE-Cut is primarily APL-ND with some specific parts under APL-SA (not clear from the outside what) and I'm not sure how you got your UH-60M base texture here.

PR #694, opened 2025-04-26
Uniform Retextures

AChesheireCat:
Adds the following uniform retextures:

- CWRIII M69 uniform in Jehlici camo (+tan variant)
- CUP Crye Gen 3 uniform in MARPAT (+ alt variant)
- PTV Jacket uniform in M08 Temperate and Winter variants


lambdatiger (commented):
Same as the other PR, are the textures from somewhere we need to attribute?

Also we may want to split these out to dependencies for PTV and CWR3 in case we ever remove one or the other (or both).

addons/customGear/uniforms/CfgVehicles.hpp
```diff
@@ -0,0 +1,190 @@
...
+            "\ptv_characters\uniforms\Data\m77_co.paa",
+            "\ptv_characters\uniforms\data\usp_gloves_co.paa",
+            "",
+            ""
+        };
+        hiddenSelectionsMaterials[] = {"\ptv_characters\uniforms\data\jacket_summer.rvmat"};
+    };
+    class potato_s_g3cw_serp_m05wdlwint_v2: ptv_s_m08 {
+        author = "AMF/VityaB/Siege-A & PTV/Chesheire";
+        scope = 2;
```
PabstMirror, line 131: I think these should be scope1 but not totally sure
usually the weapon is scope2 and the man/vic thing is just scope1

addons/customGear/uniforms/config.cpp
```diff
@@ -0,0 +1,28 @@
+#include "\z\potato\addons\customGear\script_component.hpp"
+#undef COMPONENT
+#define COMPONENT customGear_uniforms
+
+class CfgPatches {
+    class ADDON {
+        units[] = {};
+        weapons[] = {
+            "csla_jehlici",
+            "clsa_tan_jehlici",
```
PabstMirror, line 10:
```suggestion
            "potato_jehlici",
            "potato_tan_jehlici",
```

lambdatiger:
> Same as the other PR, are the textures from somewhere we need to attribute?
> 
> Also we may want to split these out to dependencies for PTV and CWR3 in case we ever remove one or the other (or both).

@AChesheireCat, did you end up getting use rights in order on this one?

AChesheireCat:
> > Same as the other PR, are the textures from somewhere we need to attribute?
> > Also we may want to split these out to dependencies for PTV and CWR3 in case we ever remove one or the other (or both).
> 
> @AChesheireCat, did you end up getting use rights in order on this one?

We should be GTG - the CWRIII retextures were sourced from Jason's mod with his express permission to repack. The CUP and PTV retextures were sourced from royalty free, attribution free stock websites iirc and pieced together by me. The Heli stuff I'm working on making my own textures since it's against ADF Re-Cut wishes and license (?) to repack those.

addons/customGear/uniforms/CfgVehicles.hpp
AChesheireCat: Pretty sure you're right - fixed!

PabstMirror (approved):

PR #696, opened 2025-04-27
put orbat in arma gen

PabstMirror:


lambdatiger (approved):
I'm not familiar with POTBOT, but I think this change makes sense.

PR #698, opened 2025-04-28
Add mock blowpipe missile

PabstMirror:


lambdatiger:
Maybe move this in with the javelin in `missleGuidanceCompat\patchCWRBAF`?

addons/missileGuidanceCompat/patchCWRBAF/config.cpp
```diff
@@ -27,13 +27,25 @@ class CfgAmmo {
...
+        class ace_missileguidance: ACEGVAR(missileguidance,type_Blowpipe) {
+            enabled = 1;
+        };
+    };
 };
 
 class CfgMagazines {
     class CA_LauncherMagazine;
     class cwr3_javelin_m: CA_LauncherMagazine {
         ammo = QGVAR(M_javelin_AA);
```
lambdatiger, line 40:
```suggestion
        ammo = QGVAR(M_javelin_AA);
        ACEGVAR(missile_manpad,enabled) = 0;
```
Will need this for them to work with the Stinger PR

PabstMirror: it's now an explicit config check, also on weapon not mag
https://github.com/acemod/ACE3/pull/10830/files#diff-a107455f1cee38185ba9f752dce8335f1df99db55070d392d1d2e85445474560R32

lambdatiger: Sounds good, which PR do I need to review this one?

PabstMirror: I think just https://github.com/acemod/ACE3/pull/10887 (I guess 10893 would be neat but not at all needed)

lambdatiger (approved):
PR looks good.
The Tab mouse joystick is a bit funky/sensitive but I don't know how you can emulate the feel of a real joystick. Maybe some sort of pull to the origin, or maybe a marker.

lambdatiger:
Had the wrong branch open

PabstMirror:
might be photon messing with the motor flare but it seems to burn out too quickly?
this will be semi-wip but I'd rather just get it in now and continue to tweak

PR #701, opened 2025-05-06
tweak nolamb macro

PabstMirror:
I think arma doesn't like the `ifdef` inside the macro block?
when loading raw configs (raw pbos or  hemtt --norap)
```
ErrorMessage: File z\potato\addons\units\nolambs\CfgVehicles.hpp, line 2: '/CfgVehicles/potato_nolambs_e_rifleman.ifdef': 'c' encountered instead of '='
Application terminated intentionally
```

not really sure why, but this change seems to do the same thing and loads ok


lambdatiger (approved):
Looks good. In the future we can probably get rid of the EH since I don't think they'll have an entry point for LAMBS, but for now we can hold onto it to be sure.

PR #703, opened 2025-05-07
Mod DUI Nametags to let users show how they want to be called

PabstMirror:
kinda hack dui nametags to add our own stuff

### DIFF
--- a/addons/dui_goby/functions/fnc_loop.sqf
+++ b/addons/dui_goby/functions/fnc_loop.sqf
@@ -23,7 +23,7 @@ if ((GVAR(show) == 0) || {GVAR(show) == 1 && {!EGVAR(safeStart,safeStartEnabled)
         // We inject the group and then a fake </t> and new line to add a third line to the nametag
 
         private _group = groupId (group _x);
-        private _formating = format ["<t font='%1' color='%2' size='%3' shadow='%4'>", diwako_dui_nametags_fontGroup, "#FF5B00", (profileNamespace getVariable ['igui_diwako_dui_nametags_h', pixelH * 64]) * 9, diwako_dui_nametags_groupFontShadow];
+        private _formating = format ["<t font='%1' color='%2' size='%3' shadow='%4'>", diwako_dui_nametags_fontGroup, GVAR(colorHex), (profileNamespace getVariable ['igui_diwako_dui_nametags_h', pixelH * 64]) * 9, diwako_dui_nametags_groupFontShadow];
         private _customGroup = format ["%1</t><br/>%2[ %3 ]", _group, _formating, _goby];
         _x setVariable ["diwako_dui_nametags_customGroup", _customGroup];
     };
--- a/addons/dui_goby/functions/fnc_loop.sqf
+++ b/addons/dui_goby/functions/fnc_loop.sqf
@@ -2,12 +2,11 @@
 
 [FUNC(loop), [], 5] call CBA_fnc_waitAndExecute;
 
-if ((GVAR(show) == 0) || {GVAR(show == 1) && {!EGVAR(safeStart,safeStartEnabled)}}) exitWith {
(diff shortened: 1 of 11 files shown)


addons/dui_goby/functions/fnc_loop.sqf
```diff
@@ -0,0 +1,30 @@
...
+        _x setVariable ["diwako_dui_nametags_customGroup", nil];
+    } else {
+        GVAR(setOn) pushBackUnique _x;
+        TRACE_2("setting",_x, _goby);
+
+        // DUI is expecting a simple string for customGroup ie `<t>%1</t>`
+        // We inject the group and then a fake </t> and new line to add a third line to the nametag
+
+        private _group = groupId (group _x);
+        private _formating = format ["<t font='%1' color='%2' size='%3' shadow='%4'>", diwako_dui_nametags_fontGroup, "#FF5B00", (profileNamespace getVariable ['igui_diwako_dui_nametags_h', pixelH * 64]) * 9, diwako_dui_nametags_groupFontShadow];
```
PabstMirror, line 26: can add settings for color, font and size like normal dui if desired
(maybe allow the tag's owner to pick the color?)

If you run a extra large font for both name and group then it might be possible to run out of vertical space, but for normal settings it should be fine (ref normal name is 10, normal group is 8, this is 9)

lambdatiger: You could always use one of the DUI group color, or a new setting is fine. Either way may need to be able to be changed to avoid issues with colorblind contrast or  other poor contrast (i.e., Bucks blown out saturation probably means orange doesn't contrast well).

kil0byt3: a more faded grey/white might be ideal so it's not too bright/overwhelming.

ideally if user-defined color settings are available, it changes the colors on your client only (not sure if possible)

having each user choose their own color would make things inconsistent, I think I prefer consistency over customization in this instance

lambdatiger: > ideally if user-defined color settings are available, it changes the colors on your client only (not sure if possible)
DUI already has 3-4 different client side colors. 

> having each user choose their own color would make things inconsistent, I think I prefer consistency over customization in this instance
They're client-side, so we're not setting and sending colors across the network (unless there's something I'm missing).

kil0byt3: gotcha, pabst saying "(maybe allow the tag's owner to pick the color?)" made me think it was choosing the color that would be displayed to others

addons/dui_goby/functions/fnc_loop.sqf
```diff
@@ -0,0 +1,30 @@
...
+    GVAR(setOn) = [];
+};
+
+{
+    private _goby = _x getVariable [QGVAR(goBy), ""];
+    if (_goby == "") then {
+        _x setVariable ["diwako_dui_nametags_customGroup", nil];
+    } else {
+        GVAR(setOn) pushBackUnique _x;
+        TRACE_2("setting",_x, _goby);
```
lambdatiger, line 20:
```suggestion
        TRACE_2("setting",_x,_goby);
```

addons/dui_goby/functions/fnc_loop.sqf
lambdatiger:
```suggestion
        private _color = [diwako_dui_nametags_groupNameOtherGroupColor, diwako_dui_nametags_groupColor] select ((group _x) isEqualTo (group _player));
        private _formating = format ["<t font='%1' color='%2' size='%3' shadow='%4'>", diwako_dui_nametags_fontGroup, _color, (profileNamespace getVariable ['igui_diwako_dui_nametags_h', pixelH * 64]) * diwako_dui_nametags_fontGroupNameSize, diwako_dui_nametags_groupFontShadow];
```
Maybe something like this?

lambdatiger (commented):
Looks fine, works reasonably well.

PR #704, opened 2025-05-08
Briefing and Markers - Fix Orbat Sorting and Catch Bad Map Icon String

lambdatiger:
This PR will:

- Allow the markers system to better handle incorrect icon strings
- Fix a few more orbat sorting issues


lambdatiger:
Fucked branch, branched from a bad git version,

PR #719, opened 2025-05-29
miscFixes - GM Update ACE Frag Compat

lambdatiger:
This PR adds the frag compats for the new M56 grenade and 120mm mortar shells


PabstMirror (approved):

PabstMirror:
maybe these should go in ace as well?

PR #720, opened 2025-05-29
miscFixes - GM Update - Change 9P133s Missiles to 9M14P to Align with the NVA & PPA Versions

lambdatiger:
This PR changes the default ATGM for the GM 9P133s to a 9M14P instead of a 9M14M as the NVA used 9M14Ps for this platform. Still allows custom magazine selection for 9M14M instead of P.


lambdatiger:
Unsure if we want to include this in ACE, lmk if you think it'd be a decent idea.

PabstMirror (approved):
sounds good to me

PR #731, opened 2025-06-05
DUI Compass Settings Update

AChesheireCat:
- Sets the default visual range of the DUI compass to the current max range (50m)

- Enables the Vehicle Compass again

- Sets the Vehicle Crew compass scale, fixing the previous issue of giant icons in the DUI hud when in a vehicle - this one might need more testing with various UI scales to verify?

- Increases the distance warning from 3m to 5m

The only thing that I really will advocate for is the compass range stuff, everything else was just something I discovered by coincidence when testing and am fine with splitting them up if we don't like the other stuff being on.

### DIFF
--- a/addons/settings/BW_Settings.inc.sqf
+++ b/addons/settings/BW_Settings.inc.sqf
@@ -108,7 +108,6 @@ if (["diwako_dui_main"] call ACEFUNC(common,isModLoaded)) then { _settings appen
 ["diwako_dui_indicators_range", 20], // User-Setting
 
 // DUI - Squad Radar - Radar
-["diwako_dui_compassRange", 50, true], // User-Setting (this one sets the starting range)
 ["diwako_dui_compassRangeLimit", 50, true], // this one sets the MAX limit
 ["diwako_dui_radar_compassrangecrew", 75, true], // User-Setting
 ["diwako_dui_enable_compass_dir", 1, true], // User-Setting "Show Bearing = only in vehicles"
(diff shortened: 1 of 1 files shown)


addons/settings/BW_Settings.inc.sqf
```diff
@@ -108,11 +108,14 @@ if (["diwako_dui_main"] call ACEFUNC(common,isModLoaded)) then { _settings appen
 ["diwako_dui_indicators_range", 20], // User-Setting
 
 // DUI - Squad Radar - Radar
-["diwako_dui_compassRange", 35, true], // User-Setting
+["diwako_dui_compassRange", 50, true], // User-Setting (this one sets the starting range)
```
PabstMirror, line 111: maybe we should not set and force this and leave it a user setting?
the `limit` setting is doing what we wanted

PabstMirror (approved):

addons/settings/BW_Settings.inc.sqf
AChesheireCat: fix'd!

PabstMirror (approved):

kil0byt3 (approved):

kil0byt3 (approved):

addons/settings/BW_Settings.inc.sqf
```diff
@@ -108,11 +108,14 @@ if (["diwako_dui_main"] call ACEFUNC(common,isModLoaded)) then { _settings appen
 ["diwako_dui_indicators_range", 20], // User-Setting
 
 // DUI - Squad Radar - Radar
-["diwako_dui_compassRange", 35, true], // User-Setting
+["diwako_dui_compassRange", 50], // User-Setting (this one sets the starting range)
```
PabstMirror, line 111: because it's a user setting (that isn't forced) this ends up not doing anything
limitation of the bw setting / CBA; the client's setting with have priority

we could one-time manually set it for them?

PR #737, opened 2025-06-22
RHS T-72 Ammo Buffs

AChesheireCat:
- Increases caliber and adds hit values to RHS T-72/T-90 rounds comparable to CUP values for same/similar rounds

- In testing, this allowed for RHS tanks to actually penetrate and kill CUP/CWR3 M60A3s and in some cases (3BM46 and 69) penetrate CUP/CWR3 M1A1s

- Rounds might be a little overtuned due to the submunitions also having hit values, but imo this at least makes RHS tanks usable against certain CUP assets

### DIFF
--- a/addons/miscFixes/patchRHSAFRF/CfgAmmo.hpp
+++ b/addons/miscFixes/patchRHSAFRF/CfgAmmo.hpp
@@ -36,30 +36,25 @@ class CfgAmmo {
     
     // Base Classes
     class Sh_125mm_APFSDS;
-    class rhs_ammo_bm_base: Sh_125mm_APFSDS {};
-    class rhs_ammo_3bm_base: rhs_ammo_bm_base {};
+    class rhs_ammo_ap_penetrator: Sh_125mm_APFSDS {};
     
     // 3BM42 - used by T-72B (original)
-    class rhs_ammo_3bm42: rhs_ammo_3bm_base {
-        caliber = 19.2157; // was 0.1
-        hit = 520; // was nil
+    class rhs_ammo_3bm42_penetrator: rhs_ammo_ap_penetrator {
+        hit = 520; // was 250
     };
     
     // 3BM42M "Mango" - used by T-72B (1985)
-    class rhs_ammo_3bm42m: rhs_ammo_3bm_base {
-        caliber = 24.7619; // was 0.1
-        hit = 542.2; // was nil
+    class rhs_ammo_3bm42m_penetrator: rhs_ammo_ap_penetrator {
+        hit = 542.2; // was 270
     };
     
-    // 3BM46 - used by T-72B3
-    class rhs_ammo_3bm46: rhs_ammo_3bm_base {
-        caliber = 26.7059; // was 0.1
-        hit = 626.8; // was nil
+    // 3BM46 - used by T-72B3  
+    class rhs_ammo_3bm46_penetrator: rhs_ammo_ap_penetrator {
+        hit = 626.8; // was 300
     };
     
-    // 3BM69 - used by T90
-    class rhs_ammo_3bm69: rhs_ammo_3bm_base {
-        caliber = 33.6842; // was 0.1
-        hit = 858.16; // was nil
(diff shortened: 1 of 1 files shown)


addons/miscFixes/patchRHSAFRF/CfgAmmo.hpp
```diff
@@ -33,4 +33,33 @@ class CfgAmmo {
...
+    
+    // 3BM46 - used by T-72B3
+    class rhs_ammo_3bm46: rhs_ammo_3bm_base {
+        caliber = 26.7059; // was 0.1
+        hit = 626.8; // was nil
+    };
+    
+    // 3BM69 - used by T90
+    class rhs_ammo_3bm69: rhs_ammo_3bm_base {
+        caliber = 33.6842; // was 0.1
```
TheCandianVendingMachine, line 62: i assume you did, but i just wanna make sure: are these rounds similar to the ones that CUP has? iirc RHS uses submunitions to penetrate, so these main rounds are meant to do nothing (but vehicle damage probably eats them in that case)

AChesheireCat: The caliber and hit values are similar (and in most cases copied 1:1). You're bang on with the vehicle damage eating the submunitions or something similar - in local testing, the new values allowed for penetration and subsequent damage to the vehicles that previously were impervious (ie. CUP/CWR3 M60A3).

TheCandianVendingMachine (approved):

addons/miscFixes/patchRHSAFRF/CfgAmmo.hpp
AChesheireCat: Are you sure? The listed values for the RHSAFRF 3BM69 penetrator (rhs_ammo_3bm69_penetrator) in the files is 

`
                hit = 300;
		typicalSpeed = 2050;
		caliber = 33.6842;
`

Am I looking at the wrong thing here?

AChesheireCat: That was actually the first thing I tried - it resulted in crew kills, but 0 damage being done to the vehicle (literally, would kill like, the gunner only and result in 0 engine dmg or hull dmg). Unless they need to be cranked to ridiculous damage levels, this seems like a better solution tbh

lambdatiger: If there's no damage applied, then there may be another underlying issue that should be addressed instead.

TheCandianVendingMachine: i think that VD could be eating the pentrator as a way to avoid spam damage. ill investigate

TheCandianVendingMachine: tweaked in https://github.com/acemod/ACE3/pull/10979. Please test and tweak this PR based on it

in testing, this lets CUP vs CUP be better. The Leopard 2s6 can actually be killed from front-shots with a T-90 now. Likewise, it lets M60 die better. However, RHS vs CUP is still iffy, penetrators may be a bit weaker

lambdatiger: I was wrong on the damage earlier.

Playing around with it, it seems ok without this PR and with the ACE PR, but if we did want to bring it on par with CUP we should match the RHS penetrators to the CUP rounds. My only worry with increased damage is it's more likely to make vehicle go catastrophic and I enjoy when they don't just detonate.

lambdatiger: I would say a start would be moving damage back to the penetrators and matching them to CUP and then if that isn't satisfactory you can tweak. Because I do like the idea of having the same round do the same damage.

PabstMirror:
Any thoughts on how this is working with the ace changes?
Or should we wait for the AP changes as well?

lambdatiger (approved):

PR #760, opened 2025-07-12
Burst Fire M4 Variants

AChesheireCat:
Adds several 3-round-burst M4 variants based on the CUP weapons. Good for era-accurate loadouts, ie. USMC c. 2012 - 2016 (they never adopted the M4A1 officially for whatever reason).

It mostly just copies the existing configs while using the M16A2 Carbine model and the M4A1 model.

- M4 Carbine
- M4 Carbine M203

- M4 Carbine RIS
- M4 Carbine RIS (Grip)
- M4 Carbine RIS M203


AChesheireCat:
@lambdatiger Yeah, sure - I just went with the M16 soundset since the M16A2 uses it lol

Same for the classnames, that's fine too

And not really, the CUP M16 inheritances were just super messy so instead of trying to figure it out, I just inherited the base class and applied select parts of the config. Seriously, the inheritances looked super messy lol

PR #763, opened 2025-07-24
M249 Recoil Re-Adjustment

AChesheireCat:
Adjusts the previous CUP M249 recoil adjustment to be tuned down following [feedback in the discord.](https://discord.com/channels/204621032428929025/1393171695401173083)

This changes the recoil from a significant permanent up-and-to-the-right recoil to a much softer temporary "bouncy" recoil. The intent is to reduce the need to control the recoil with the mouse, but still prevent the M249 from being a laser beam @ 100m.

### DIFF
--- a/addons/miscFixes/patchCUP/config.cpp
+++ b/addons/miscFixes/patchCUP/config.cpp
@@ -244,28 +244,22 @@ class CfgRecoils {
         permanent = 0.1;
         temporary = 0.005;
     };
-    class GVAR(recoil_SAW): recoil_default { // original modified saw recoil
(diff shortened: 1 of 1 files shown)


AChesheireCat:
Notes for further consideration: Maybe tune down the first shot recoil slightly? In iron sights, it seems to jump enough to possibly disorient, but is fine for versions with optics capability.

lambdatiger:
Where you thinking of changing prone recoil, or are you happy with it?

PabstMirror (approved):

PR #774, opened 2025-09-02
CUP MLRS Submunitions Fix

AChesheireCat:
"Nerfs" the CUP GRAD, Hilux MLRS, Hilux UB-32 launcher damage by removing the inherited 230mm submunition from the warhead. This makes the damage much more reasonable instead of being portable nukes.

Additionally, buffs the Hilux MLRS and UB-32 accuracy so that way it's reasonably usable for grid-square targets post-damage nerf.


AChesheireCat:
Ah shoot, forgot to add credit to martin509 over at the CUP discord for the bug report and subsequent fix.

PabstMirror (approved):

PR #784, opened 2025-09-26
Spectate - Show outgoing wounds

PabstMirror:


lambdatiger:
Is this supposed to catch outgoing wounds only done to other players? Testing with AI I only get kills in the feed.

PabstMirror:
AI won't have a kill track feed to check
so it will only show other players you wounded

I need to estimate how expensive this is, it's only for spectate but I still don't want it to cause stutter

lambdatiger:
So it's moving to just being a TvT feature?

lambdatiger:
> So it's moving to just being a TvT feature?

It's fine if it is, if so this should be fine now.

PabstMirror:
mainly tvt, but violence can happen anytime

lambdatiger (approved):

PR #791, opened 2025-10-02
vehicleCoasting - Forget Targets on Bail

lambdatiger:
This PR forces AI bailing out from vehicles to forget all target info within the last ten seconds. The goal is to stop AI crew from popping out of vehicles and simultaneously spraying down players.


lambdatiger:
This one could be neater. Let me know if we prefer having them remember some amount of target info, just maybe not accurate.

PR #815, opened 2025-11-17
adminMenu - Add All Win/Lose End Mission Buttons

lambdatiger:
Adds buttons so everyone can get a win or lose message.


addons/adminMenu/functions/fnc_endMission.sqf
PabstMirror: need to peek at func, but this might expect a side type

PR #817, opened 2025-11-22
Various - Cleanup Headers

PabstMirror:


addons/markers/functions/fnc_transferMarker.sqf
```diff
@@ -36,7 +36,7 @@ private _newPosATL = getPosATL _newObject;
 if (_newPosATL isNotEqualTo [0, 0, 0]) then {
     _markerArray set [5, _newPosATL];
 };
-if !(isNull _newSide) then {
+if (_newSide isNotEqualTo sideUnknown) then {
```
PabstMirror: this is the only code change, 
but nothing was calling with 3rd arg so I don't think this will have any effects?

lambdatiger: Looks good!

PR #818, opened 2025-11-29
CUP AK Tweaks

AChesheireCat:
- Corrects AK-74 and other 5.45 derivative rifles' fire-rate to the IRL 652 RPM (was previously 600 RPM).

- Adds custom recoil defines seeking to emulate GM's AKM recoil

- Adds duplicate CUP AK (7.62) rifles that utilize the new custom recoil - this was originally intended to add more differentiation between the 5.45 rifles and the 7.62 rifles as well as a little bit more TvT balancing.

### DIFF
--- a/addons/miscFixes/patchCUP/config.cpp
+++ b/addons/miscFixes/patchCUP/config.cpp
@@ -417,33 +417,33 @@ class CfgWeapons {
     // AKM
     class potato_arifle_AKM: CUP_arifle_AKM_Early { // CUP AKM using custom recoil
         scope = 1; 
-        scopeArsenal = 2; // was 0
+        scopeArsenal = 1;
         recoil = QGVAR(recoil_akm);
     };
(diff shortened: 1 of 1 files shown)


lambdatiger:
Is there any easy way to tell the difference between these AKs and the vanilla ones Potato has edited?

lambdatiger (changes requested):

lambdatiger:
> > Do you mind hiding them (`scopeArsenal = 1`) for a update rotation, and if players like it we switch full time (hide/change base CUP). Maybe we get gunner to use them for a few missions as well?
> 
> Not at all, that was kinda the original intention since I initially made these just for me. The only reason why I didn't set them to hidden by default was because I noticed that with the hidden COOP M249s, hiding them in the arsenal seems to just preclude them from use and if anyone else was interested in them, I'd want them to have easy access to them.

Part of that is because I didn't make an effort to have people check them, or even use them myself yet. I'll ask Gunner and if he would be down and we can make that change then.

PabstMirror (approved):

PR #822, opened 2025-12-03
briefing - add more tabs

TheCandianVendingMachine:
intent and enemy info tabs are now present for all sides


lambdatiger:
Handled the `adminMenu` version

PR #824, opened 2025-12-05
Core - Add arsenal stat for recoil

PabstMirror:


lambdatiger (approved):
This feels like a better rep of recoil looking through.

PR #861, opened 2026-03-31
Add more mission types, enhance mission attribute timers

TheCandianVendingMachine:
title.

This was stolen from a grander WIP change with mission tags, but that is taking long and I want this now


lambdatiger (changes requested):
I don't like getting rid of an unset/error state so we can prompt mission makers to select one. 

In addition, these changes also affect:

- adminMenu - `fnc_endMission.sqf` - lines 7-8
- extension - `fnc_safeStartMessages.sqf` - lines 53-55
- missionMaking - `fnc_onMissionSave.sqf` - lines 118-120
- safeStart - `fnc_initForceEndSafeStartsqf` - lines 30-31

The above need to be addressed before we merge. I've already begun addressing changes to safeStart - `fnc_fillSafeStartEquip.sqf` to avoid future merge conflcits

addons/adminMenu/functions/fnc_endMission.sqf
```diff
@@ -1,11 +1,12 @@
 #include "script_component.hpp"
+#include "\z\potato\missionTesting\script_macros.hpp"
```
lambdatiger:
```suggestion
#include "\z\potato\addons\missionTesting\script_macros.hpp"
```

addons/extension/functions/fnc_safeStartMessages.sqf
```diff
@@ -1,4 +1,5 @@
 #include "..\script_component.hpp"
+#include "\z\potato\missionTesting\script_macros.hpp"
```
lambdatiger:
```suggestion
#include "\z\potato\addons\missionTesting\script_macros.hpp"
```

addons/editorExtensions/functions/fnc_sliderTime_load.sqf
```diff
@@ -0,0 +1,17 @@
+#include "script_component.hpp"
+
+private _increment = getNumber (_config >> "increment");
+private _maxTimeSeconds = getNumber (_config >> "maxTimeSeconds");
```
lambdatiger, line 5: `_config` isn't defined in this file nor in the handler call. I don't know of any vanilla EH's that define variables like this for you, if they do, add a hemtt variable warning ignore.

lambdatiger (changes requested):
Looks better, few issues still.

lambdatiger:
Do you want to handle your merge conflict or are you ok with anyone else doing it?

addons/editorExtensions/functions/fnc_sliderTime_load.sqf
lambdatiger: `_value` later in this function as well.

TheCandianVendingMachine: idk what the fuck is the reason why these are defined, but vanilla elements these are based on use the magic variables too

lambdatiger: I love arma.
`//IGNORE_PRIVATE_WARNING["_config", "_value"];` at the top then.

TheCandianVendingMachine:
Someone else can, this isn’t the most conflictful change

lambdatiger (changes requested):
Currently `potato_missiontesting_missionTimeLength` and `potato_missiontesting_SSTimeGiven` are in minutes, this PR changes them to be in seconds. It'd be easier to save the value in minutes than it would be to change all the locations that use it, but up to you if you want to track down all the uses and update them.

Other than that, it works.

lambdatiger:
Quick patch, skipped extension since it isn't time sensitive and defaults to 9 minutes.

PabstMirror (approved):

lambdatiger (approved):

PR #862, opened 2026-04-02
Radios/Markers - Try to wait for group data to be ready

PabstMirror:
not tested


lambdatiger (approved):
Can't seem to fully test it (running into overrun time), but basic functionality seems to still work.

PR #863, opened 2026-04-03
Various - Update Fixes

lambdatiger:
This PR:

- Added a hashmap to allow factions to indicate base loadout setting to allow for future changes to loadout path.
- Fix mission testing menu type issue.
- Set's goBy to default to always show (still changeable in settings)
- Disable ACRE `acre_sys_radio_edenSetup`
- Update safe start menu for future mission type settings

### DIFF
--- a/addons/radios/CfgEden.hpp
+++ b/addons/radios/CfgEden.hpp
@@ -341,8 +341,8 @@ class Cfg3DEN {
             class acre_attributes {
                 class Attributes {
                     class acre_sys_radio_edenSetup {
+                        condition = 0;
                         defaultValue = "[]";
-                        expression = "";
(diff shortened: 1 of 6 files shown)


PabstMirror (approved):

addons/radios/CfgEden.hpp
```diff
@@ -338,6 +338,14 @@ class Cfg3DEN {
                     };
                 };
             };
+            class acre_attributes {
+                class Attributes {
+                    class acre_sys_radio_edenSetup {
+                        condition = 0;
```
PabstMirror, line 344: we could also `delete` it, but I don't think that would be any better

PR #864, opened 2026-04-08
GoBy - Update to use `dui_nametags_customInfo`

PabstMirror:


lambdatiger (approved):

lambdatiger (approved):
Nice fix with the colors.

PabstMirror:
ref https://github.com/diwako/diwako_dui/issues/275
but won't actually change a whole lot for this

it will be a little confusing because text size will be on dui settings and text color will be back in potato settings,
I guess we could restore the text size back in potato and just ignore all the custom_x settings in dui?

lambdatiger:
Up to you, either way it's a bit split since the DUI settings are separate. I'm happy if we keep ours as well and just ignore DUI's setting.

lambdatiger (approved):

PR #866, opened 2026-04-09
arsenalExport - Partial Loadout Sanitizing and Potato GP-25 Rounds

lambdatiger:
This PR removes the "NotSet" strings for MMG/HMG/MAT/HAT/Sniper/Spotter/Pistol/SMG from the arsenal export if a mission maker does not define them. In addition, it defaults to CUP GP-25 rounds for compatible rifles first.

### DIFF
--- a/addons/arsenalExport/functions/fnc_export.sqf
+++ b/addons/arsenalExport/functions/fnc_export.sqf
@@ -91,9 +91,9 @@ private _glMuzzle = (getArray (configFile >> "CfgWeapons" >> GVAR(loadout_glrifl
 private _glMags = [configFile >> "CfgWeapons" >> GVAR(loadout_glrifle) >> _glMuzzle] call CBA_fnc_compatibleMagazines;
 switch (true) do {
     case (({"CUP_1Rnd_HE_GP25_M" == _x} count _glMags) > 0): {
-        _lines pushBack format ['#define GLRIFLE_MAG_SMOKE "CUP_1Rnd_SMOKE_GP25_M:2","CUP_1Rnd_SmokeRed_GP25_M:2"'];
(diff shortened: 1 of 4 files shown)


TheCandianVendingMachine:
We intentionally moved away from the gp25 rounds. i’d rather create our own with the same specs as the default gp and copy the model

lambdatiger:
> We intentionally moved away from the gp25 rounds. i’d rather create our own with the same specs as the default gp and copy the model

Where was that discussion? It was there last time I edited this file, but in the wrong order so it'd usually never trigger, I'm actually a bit confused since it wasn't mentioned in any PR as to removing it.

My main thought is that it's weird to pick up standard 40mm from EI firing GP-25s.

TheCandianVendingMachine:
Oh this was ages ago. if this was made without fuss keep it. I think we made the change to ensure ammo compatibility between every gun, it was an AACO thing

lambdatiger:
> Oh this was ages ago. if this was made without fuss keep it. I think we made the change to ensure ammo compatibility between every gun, it was an AACO thing

So I removed this tidbit last time I committed to this file (https://github.com/BourbonWarfare/POTATO/commit/6f9df76384efe27eb992c8e3dd5a3700abbd0e79#diff-973b22baddc5e04453eebf30760ae88adf87cdab44059072b1c483f8b83bec44) because it would never get hit and there were some issues with getting magazines wrong. Since then the backend issues were fixed. I get AACO may have had issues, but it was removed last year.

lambdatiger:
Added custom GP-25 rounds that don't rely on external mods, but still keep magazine types different.

PabstMirror (approved):

PR #879, opened 2026-05-04
Spectate - Show who verified death

PabstMirror:


lambdatiger:
Is the idea for the check to be the last before death? Or something else?

addons/spectate/XEH_postClientInit.sqf
```diff
@@ -38,4 +38,12 @@ if (GVAR(enabled) && hasInterface) then {
             [_open] call FUNC(setRespawn);
         };
     }] call CBA_fnc_addEventHandler;
+
+    ["ace_treatmentSucceded", {
+        params ["_medic", "_patient", "", "_classname"];
+        if ((_classname != "CheckResponse") || {!alive _patient} || {_medic != ace_player}) exitWith {};
```
PabstMirror:
```suggestion
        if ((_classname != "CheckResponse") || {alive _patient} || {_medic != ace_player}) exitWith {};
```

lambdatiger, line 44:
```suggestion
        if ((alive _patient) || {_classname != "CheckResponse"} || {_medic != ace_player}) exitWith {};
```

lambdatiger: And do you want it to be the last to check response, or first? Could check if the variable was set and exit if you want that.

lambdatiger (approved):

addons/spectate/XEH_postClientInit.sqf
```diff
@@ -38,4 +38,12 @@ if (GVAR(enabled) && hasInterface) then {
             [_open] call FUNC(setRespawn);
         };
     }] call CBA_fnc_addEventHandler;
+
+    ["ace_treatmentSucceded", {
+        params ["_medic", "_patient", "", "_classname"];
+        if ((_classname != "CheckResponse") || {alive _patient} || {_medic != ace_player}) exitWith {};
```
lambdatiger, line 44: Only feedback is this will update again and again as each person checks if that's a problem

lambdatiger (approved):

PR #886, opened 2026-05-13
adminMenu - Add Stabalize Group Button

lambdatiger:
This PR adds a "Stabilize Group" button to the right of the stabilize player button.
image


PabstMirror (approved):
🔪

PR #888, opened 2026-05-19
core - add fix for `acre_sys_radio_setup` errors

PabstMirror:
can't put in potato_radios because that runs after acre_sys_radios because of requiredAddons load order
I'm not sure if potato_core is guaranteed to run first, but this seems to fix the issue

### DIFF
--- a/addons/core/XEH_postInit.sqf
+++ b/addons/core/XEH_postInit.sqf
@@ -1,5 +1,19 @@
 #include "script_component.hpp"
 
+// this is in core because potato_radios will run after acre
+if (isServer) then {
+    private _list = allUnits select { !((_x getVariable ["acre_sys_radio_setup", ""]) isEqualType "") } apply { [_x, vehicleVarName _x] };
+    [{
+        {
+            _x params ["_unit", "_varName"];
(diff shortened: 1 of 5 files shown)


addons/core/XEH_postInit.sqf
```diff
@@ -1,5 +1,19 @@
 #include "script_component.hpp"
 
+// this is in core because potato_radios will run after acre
+if (isServer) then {
+    private _list = allUnits select { !((_x getVariable ["acre_sys_radio_setup", ""]) isEqualType "") } apply { [_x, vehicleVarName _x] };
+    [{
+        {
+            _x params ["_unit", "_varName"];
+            ["potato_adminMsg", [format ["Warning: Bad acre var on slot %1 (%2)", _varName, name _unit]]] call CBA_fnc_globalEvent;
```
PabstMirror: I don't think we actually need this, and really don't care if the vars are set if we are just clearing them anyways?

lambdatiger: Sounds fine, do we still want to log it? Also do we want to clear the attribute on save so if people have an older version of bwmf it cleans itself?

lambdatiger (approved):

PR #892, opened 2026-05-29
various - Spectate Update and New CSW Magazines 

lambdatiger:
This PR
- Disable players reporting their FPS to the server if a MM does not enable it (previously inaccessible unless turned on)
- Adds a 32 round 40x53mm CSW magazine for Potato ammos
- Updates the spectate death drawing system to be a bit visually cleaner.

### DIFF
--- a/addons/zeusUtils/XEH_postInit.sqf
+++ b/addons/zeusUtils/XEH_postInit.sqf
@@ -11,10 +11,3 @@ if (isServer && isMultiplayer) then {
 
 if !(hasInterface) exitWith {};
 #include "initKeybinds.inc.sqf"
-
-["CBA_SettingsInitialized", {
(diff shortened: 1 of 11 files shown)


addons/zeusUtils/XEH_postInit.sqf
```diff
@@ -14,7 +14,7 @@ if !(hasInterface) exitWith {};
 
 ["CBA_SettingsInitialized", {
     TRACE_4("CBA_SettingsInitialized EH Client",_this,GVAR(fpsDisplayEH),GVAR(fpsAvgCalcEH),GVAR(fpsAvgCalc));
-    if (isMultiplayer) then {
+    if (isMultiplayer && GVAR(missionFPSEnable)) then {
         [] call FUNC(initLocalFPSEH);
```
PabstMirror: we add this is 2 places, the setting init and here, so I think we could just remove this one

lambdatiger: Resolved

PabstMirror (approved):

PR #895, opened 2026-06-04
customGear - Grenade Launched Flashbang Round

lambdatiger:
This PR adds adds a GL flashbang round.

TODO
- [x] Test it

### DIFF
--- a/addons/customGear/CfgAmmo.hpp
+++ b/addons/customGear/CfgAmmo.hpp
@@ -115,7 +115,7 @@ class CfgAmmo {
         displayNameshort = "Flash";
         displayName = "40x46mm Flash";
         // Damage - reduced for smaller one
-        explosive = 0;
+        explosive = 1E-7;
         hit = 15;
         indirectHitRange = 0;
         maxSpeed = 250;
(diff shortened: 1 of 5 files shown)


addons/customGear/CfgAmmo.hpp
```diff
@@ -98,6 +98,35 @@ class CfgAmmo {
         soundsetexplosion[] = {};
     };
 
+    // Flashbank ammo
+    class AMMO(40x46mm_FLASH): AMMO(40x46mm_HEDP_M433) {
+        // flashbang info
+        ACEGVAR(grenades,flashbang) = 1;
```
PabstMirror: https://github.com/acemod/ACE3/blob/master/addons/grenades/functions/fnc_throwGrenade.sqf#L21

only throws

might be able to use ammo's explosion event handler and call `ace_grenades_fnc_flashbangThrownFuze`

lambdatiger: Yeah, I'll write some stuff for that, would need some locality checks, but this can go in the next update. Don't have it in me tonight.

lambdatiger:
Looks good to me, seems to work. Unsure about networking with the new ACE method.

PabstMirror (approved):

PR #897, opened 2026-06-06
markers - Replace Unit Only with Occluded Version

lambdatiger:
This PR replaces the "unit only" setting with a occlusion-fading method for intra-group markers.

### DIFF
--- a/addons/markers/initSettings.inc.sqf
+++ b/addons/markers/initSettings.inc.sqf
@@ -25,7 +25,7 @@
     false, // default value
     true, // isGlobal
     {
-        if (!isNil QGVAR(intraAlphaPFEH) && {GVAR(intraAlphaPFEH) >= 0}) then {
+        if !(isNil QGVAR(intraAlphaPFEH)) then {
             [GVAR(intraAlphaPFEH)] call CBA_fnc_removePerFrameHandler;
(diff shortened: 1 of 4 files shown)


addons/markers/initSettings.inc.sqf
```diff
@@ -19,13 +19,18 @@
 ] call CBA_fnc_addSetting;
 
 [
-    QGVAR(intraFireteam_playerOnly), "CHECKBOX",
```
PabstMirror, line 22: old var used in addons\safeStart\functions\fnc_fillSafeStartEquip.sqf

PabstMirror (approved):

addons/markers/initSettings.inc.sqf
```diff
@@ -19,13 +19,18 @@
...
-    ["intraFireteam - player only"],
+    QGVAR(intraFireteam_occlude), "CHECKBOX",
+    ["intraFireteam - Occlude Markers", "Occlusion and distance causes intra fireteam markers to fade"],
     ["POTATO - Mission Maker", "Markers"],
     false, // default value
     true, // isGlobal
```
PabstMirror, line 26: we could make this false
so people can test as they want, and then mission makers can just force enable?

PR #903, opened 2026-06-28
Respawn - Add hud icon for MSV Senior Rifleman

PabstMirror:
ref https://github.com/BourbonWarfare/bwmf/pull/216


lambdatiger (approved):

PabstMirror:
https://github.com/diwako/diwako_dui/blob/master/addons/radar/functions/fnc_getIcon.sqf#L43
just noting that this will take priority on most things
e.g. if they take actual leader of squad, it will still show just show the new icon instead of the leader (slash icon)

main motivation for this if if we do msv, and squad lead stays in the vic it can be confusing on who ftl is on the radar

PR #925, opened 2026-08-03
arsenalExport - Make CBA_DisposableLaucnhers Default to Unloaded Without Mag

lambdatiger:
This PR will replace CBA disposable loaded launchers in export with the base version, i.e., `"CUP_launch_M72A6_Loaded"` => `"CUP_launch_M72A6"`

### DIFF
--- a/addons/arsenalExport/functions/fnc_buttonClick.sqf
+++ b/addons/arsenalExport/functions/fnc_buttonClick.sqf
@@ -63,8 +63,14 @@ case ("mmg"): {
         systemChat format ["[Set %1]: %2 %3", _fncString, GVAR(loadout_mmg), GVAR(loadout_mmgMags)];
     };
 case ("at"): {
-        GVAR(loadout_at) = secondaryWeapon _unit;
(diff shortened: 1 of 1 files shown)


PabstMirror:
I need to take a look at how this will effect AI
we may just need to do a manual call to `cba_disposable_fnc_changeDisposableLauncherClass` for them
or it might already be handled

lambdatiger:
> I need to take a look at how this will effect AI we may just need to do a manual call to `cba_disposable_fnc_changeDisposableLauncherClass` for them or it might already be handled

I currently give AI the CBA disposable base class in my coops, but double check in case I just haven't noticed it failing.

lambdatiger:
Looks like it should work from [CBA disposables XEH_preInit](https://github.com/CBATeam/CBA_A3/blob/520ec5caffc6e50e435e32e825c8b2dc2509e212/addons/disposable/XEH_preInit.sqf#L13), but I might be missing something.

PabstMirror:
loadout event is players only and the `InitPost` would run before our assignGear

but this is what I was thinking of, we already handle the reversing of launchers in assignGear
(normal launcher -> loaded/ready launcher) so AI will be able to fire
https://github.com/BourbonWarfare/POTATO/blob/master/addons/assignGear/functions/fnc_getWeaponArray.sqf#L77

so everything was expecting the starting launcher already, so this change is perfect

PabstMirror (approved):

PR #926, opened 2026-08-05
Add 3den button to disable ace wind

PabstMirror:


lambdatiger:
This looks like it should work, but I can not test it.

PR #932, opened 2026-08-18
Admin Menu - Add button to fix speaker/ai callouts

PabstMirror:
it's mostly diagnostic right now
but I hope this will help

### DIFF
--- a/addons/adminMenu/XEH_postInit.sqf
+++ b/addons/adminMenu/XEH_postInit.sqf
@@ -94,3 +94,15 @@ if (isServer) then {
 if (hasInterface) then {
     [QGVAR(changeRadioSim), {call FUNC(newRadioMode)}] call CBA_fnc_addEventHandler;
 };
+
+[QGVAR(fixSpeaker), {
+    params [["_unit", objNull, [objNull]], ["_mode", "server", ["server","client"]]];
+    if (isNull _unit) exitWith {WARNING_1("Bad unit [%1] disconnect?",_unit);};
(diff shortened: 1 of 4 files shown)


addons/adminMenu/XEH_postInit.sqf
```diff
@@ -94,3 +94,15 @@ if (isServer) then {
...
     [QGVAR(changeRadioSim), {call FUNC(newRadioMode)}] call CBA_fnc_addEventHandler;
 };
+
+[QGVAR(fixSpeaker), {
+    params [["_unit", objNull, [objNull]], ["_mode", "server", ["server","client"]]];
+    if (isNull _unit) exitWith {WARNING_1("Bad unit [%1] disconnect?",_unit);};
+    if (_mode == "server") then {
+        ["potato_adminMsg", [format ["- sees [%1]", speaker _unit], "server"]] call CBA_fnc_globalEvent;
+    } else {
+        ["potato_adminMsg", [format ["- sees [%1] and attempting fix", speaker _unit], profileName]] call CBA_fnc_globalEvent;
```
lambdatiger, line 104: Probably worth changing some of these to server logging after we track down the issue, unless you think it'll all help in the moment.

lambdatiger (approved):
Should work from testing.

addons/adminMenu/XEH_postInit.sqf
PabstMirror: live data might make debugging easier the first time
can def move to rpt after it gets used

PR #938, opened 2026-08-28
mapMarkers - Add Map Marker Control System

lambdatiger:
This PR adds a module and system to restrict map markers to within a group. In addition, the module allows control to
- Disable map marker sharing (disable markers on non-group channels, except for authorized users)
- Enable/Disable sharing markers
- Enable/Disable copying markers from other people
- Enable/Disable copy from enemy
- The sharing radius

Future Plans include
- Allowing/tracking markers for taking off of dead bodies
- Zeus tools
- Live shared markers for briefing.

### DIFF
--- a/addons/mapMarkers/functions/fnc_recieveMarks.sqf
+++ b/addons/mapMarkers/functions/fnc_recieveMarks.sqf
@@ -1,0 +1,40 @@
+#include "..\script_component.hpp"
+/*
+ * Author: Lambda.Tiger
+ * This function is run by a module on init. It takes the configured
+ * radio parameters, adds relevant events and ACE actions, and configures
+ * local variables for marker sharing
+ *
+ * Arguments:
+ * 0: The module being initialized
+ * 2: Whether the module is active
+ *
+ * Examples:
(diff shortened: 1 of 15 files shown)


addons/mapMarkers/functions/fnc_recieveMarks.sqf
```diff
@@ -0,0 +1,42 @@
...
+};
+if (ace_player getSlotItemName 608 == "") exitWith {
+    ["You tried to copy someone's markers but you don't have a map.", true, 5] call ACEFUNC(common,displayText);
+};
+
+["Markers copied to map.", true, 5] call ACEFUNC(common,displayText);
+{
+    if (_y#0) then {
+        _y params ["", "_color", "_pts"];
+        private _marker = createMarkerLocal [_x, [_pts#0, _pts#1, 0]];
```
PabstMirror, line 29: super minor, i think current profiling branch throws an error on creating marker with name same as existing
we might want to think about handling copying same marks twice?

PabstMirror (approved):

PR #940, opened 2026-08-28
Addition of a Wiesel 2 HAT variant

blah2355:
This Wiesel 2 AT variant borrows the laser guided ATGM system from the Rooikat 120 (Rhino MGS). Missiles ideally lock on to laser targets but can be SACLOS guided.


addons/wieselHAT/CfgAmmo.hpp
```diff
@@ -0,0 +1,23 @@
+class cfgAmmo
+{
+	class M_120mm_cannon_ATGM_LG;
+	class POTATO_M_120mm_cannon_ATGM_LG: M_120mm_cannon_ATGM_LG
```
lambdatiger: lowercase "POTATO" for classnames

addons/wieselHAT/CfgAmmo.hpp
```diff
@@ -0,0 +1,23 @@
...
+		{
+			ascendHeight=250;
+			descendDistance=400;
+			minDistance=400;
+			ascendAngle=25;
+		};
+		class Direct
+		{
+		};
+		manualControl = 1;
```
lambdatiger: Probably want to make this ACE MG instead of vanilla, I think there's a good profile for it?

addons/wieselHAT/CfgWeapons.hpp
```diff
@@ -0,0 +1,16 @@
...
+		displayName = "120mm MARUK ATGM";
+		magazines[]=
+		{
+			"2Rnd_120mm_LG_missiles"
+		};
+		weaponLockDelay = 0;
+		weaponLockSystem = 4;
+		cmImmunity = 1;
+		canLock = 0;
+	};
```
lambdatiger: Are the particle effect memory points going to work for the same?

lambdatiger (commented):
A few thoughts:
- This could be a subaddon in `customGear`
- We generally
  - Open braces on the same line (e.g., `class TopDown {`), https://en.wikipedia.org/wiki/Indentation_style#K&R
  - Start lowercase with classnames (i.e., POTATO)
I'll take a closer look when I have a moment.

addons/wieselHAT/CfgAmmo.hpp
PabstMirror: I'm not sure if ace has a top-down mode quite like this vanilla
I guess that's a good question, with `manualControl` it's more SACLOS than laser-guided, I think?

PabstMirror (approved):

PabstMirror:
sorry for rushing some things, but we wanted to get this in for the update
thanks for the PR

PR #945, opened 2026-09-03
Various - Update some headers

PabstMirror:


addons/vehicleCoasting/functions/fnc_forgetTargets.sqf
```diff
@@ -8,8 +8,8 @@
 *
 *
 * Arguments:
-* _vehicle - A vehicle suffering a cook-off or engine fire event
-* _unit - A unit in a vehicle that should bail out.
+* 0: _vehicle - not used <OBJECT>
```
lambdatiger: Would rather we didn't lose the descriptive part.

lambdatiger:
I like this, it unifies things. Is part of what you said before about automatically generating docs?

PabstMirror:
hemtt can maybe use headers to check for function calls and some param stuff
it's not a big deal, i just took a shot at converting some over so they parse

lambdatiger (approved):

PR #948, opened 2026-09-05
safeStart - Add Mission Timer to Map

lambdatiger:
This PR adds a text box to the map that is left empty and transparent unless the mission uses BWMF, in which case it will show:
- Safe start elapsed time during safe start
- Time since the end of safe start any other time

Overhead on my machine is about 300ms every 90 minutes or about 100ms per 30 minutes (~30us per second).


addons/safeStart/XEH_postInit.sqf
```diff
@@ -43,4 +43,14 @@ if(hasInterface) then {
...
     }] call CBA_fnc_addEventHandler;
+    if (getMissionConfigValue ["bwmfDate", ""] != "") then { // non-bwmf missions don't need the timer
+        private _ctrl = (findDisplay 12) displayCtrl IDC_MAPTIME_CLOCK;
+        uiNamespace setVariable [QGVAR(missionClock), _ctrl];
+        _ctrl ctrlSetBackgroundColor [0,0,0,0.7];
+        _ctrl ctrlSetText "00:00:00";
+        _ctrl ctrlSetPositionW (ctrlTextWidth _ctrl);
+        _ctrl ctrlSetPositionH (ctrlTextHeight _ctrl);
+        _ctrl ctrlCommit 0;
+        GVAR(mapTimerID) = [FUNC(updateMapTimer), 0.25, []] call CBA_fnc_addPerFrameHandler;
```
PabstMirror: in mp findDisplay12 won't exist here, the briefing map is one of 3 different RscDisplayGetReadyX
this will use the map drawEH so no code runs when not on map, and sync second to map clock

```suggestion
        [{!isNull findDisplay 12},
        {
            private _ctrl = (findDisplay 12) displayCtrl IDC_MAPTIME_CLOCK;
            uiNamespace setVariable [QGVAR(missionClock), _ctrl];
            _ctrl ctrlSetBackgroundColor [0,0,0,0.7];
            _ctrl ctrlSetText "00:00:00";
            _ctrl ctrlSetPositionW (ctrlTextWidth _ctrl);
            _ctrl ctrlSetPositionH (ctrlTextHeight _ctrl);
            _ctrl ctrlCommit 0;
            
            GVAR(mapClockLast) = -1;
            private _map = (findDisplay 12) displayCtrl 51;
            _map ctrlAddEventHandler ["Draw", {
                private _time = floor (dayTime * 3600); // synchronize clocks
                if (_time == GVAR(mapClockLast)) exitWith {};
                call FUNC(updateMapTimer);
                GVAR(mapClockLast) = _time;
            }];
        }, []] call CBA_fnc_waitUntilAndExecute;
```

lambdatiger: I forgot about the display timing. I don't like the ctrl handler, or at least using `dayTime`. Is a good reminder to just ad an eventhandler to the control.

lambdatiger: To be clear, I'd rather they were syncopated, but accurate to the behind the scenes stuff.

lambdatiger: Everything about this is smarter than what I was going to do. Let's stick with it.

PabstMirror (approved):

addons/safeStart/XEH_postInit.sqf
```diff
@@ -43,4 +43,25 @@ if(hasInterface) then {
...
+            _ctrl ctrlSetBackgroundColor [0,0,0,0.7];
+            _ctrl ctrlSetText "00:00:00";
+            _ctrl ctrlSetPositionW (ctrlTextWidth _ctrl);
+            _ctrl ctrlSetPositionH (ctrlTextHeight _ctrl);
+            _ctrl ctrlCommit 0;
+            
+            GVAR(mapClockLast) = -1;
+            private _map = (findDisplay 12) displayCtrl 51;
+            _map ctrlAddEventHandler ["Draw", {
+                private _time = floor (dayTime * 3600); // synchronize clocks
```
PabstMirror, line 60: I think depending on fps and if time offset is something unlucky like 0.98
it might skip by 2 seconds sometimes

we can just wait and see if anyone has problems

PR #951, opened 2026-09-06
AdminMenu - Attempt to autofix bad voice after respawn

PabstMirror:


lambdatiger (approved):
Looks fine. Do you want the extra spam on reinforce in admin messages or server logging?

PR #952, opened 2026-09-07
recruits - Add whitelist for member tags

TheCandianVendingMachine:
In progress until a decision is made if we 1) want this 2) want to exclude in-game recruit markers


lambdatiger (approved):
Only thought is depending on how gracefully we shutdown the servers for restarts, we may need a `saveProfileNamespace` on mission end.

PR #954, opened 2026-09-08
Recruits - Have server check if member

PabstMirror:
ref https://github.com/BourbonWarfare/potato_server/pull/2


TheCandianVendingMachine (approved):

addons/recruits/functions/fnc_toggleRecruitInfo.sqf
```diff
@@ -17,8 +17,6 @@ TRACE_1("Params",_this);
 
 params ["_enable"];
 
-if ([player] call FUNC(isNotMember)) exitWith { LOG("Player not in squad.xml, exiting"); };
```
PabstMirror, line 20: just noting: this means everyone runs the pfed
allows isNotMember to change mid-session to handle any possible network delays
specifically it could switch from false->true

